PR #21, opened 2020-02-19
improveing naming and documentation in tensor product section

bkmi:
I changed the confusing name mixing_matrix into permutation_matrix which is well defined mathematically. I also added documentation for the cases which I could understand including usage and doctest.

This isn't quite done yet, there is one confusing function left which I still don't exactly understand...


e3nn/SO3.py
```diff
@@ -376,13 +406,13 @@ def tensor_productRs(Rs_1, Rs_2, get_l_output=selection_rule):
...
                 index_out += dim_out
 
             index_2 += dim_2
         index_1 += dim_1
 
-    return Rs_out, mixing_matrix
+    return Rs_out, Q
 
 
 def elementwise_tensor_productRs(Rs_1, Rs_2, get_l_output=selection_rule):
```
bkmi: This function needs clarifying. Is it also computing an orthonormal change of basis? Could we also get a usage case?

blondegeek:
The mixing matrix in tensor product is actually a Clebsch-Gordon matrix,
not a permutation matrix.

e3nn/SO3.py
mariogeiger: Yes it is an orthogonal change of basis!

`tensor_product = einsum('ijk,j,k->i', tensor_productRs(Rs1, Rs2)[1], input1, input2)`

mariogeiger:
This is very nice, thanks a lot @bkmi 
I wait that you add the needed clarification that you talk about and I merge when you are ready

e3nn/SO3.py
blondegeek: I disagree with calling it a change of basis only because typically that is interpreted as a rotation `einsum('ij,j->i', ...)`. Rather, it is a contraction of two indices into one index. You cannot turn `ki,i->k` independently -- you must contract over both `i` and `j` simultaneously. Perhaps we can just call it `clebsch_gordan_tensor`?

blondegeek: Further clarification -- the `ijk` tensor in `ijk,j,k->i` is not a change of basis, but if you go from `ijk,j->ik` and then `ik,k->i` -- this intermediate tensor `ik` is a proper change of basis tensor.

blondegeek: Gah! Sorry @bkmi -- all my comments are specifically for the tensor product and not the other mixing matrices which are indeed permutation matrices. My apologies!

bkmi: Regarding specifically about the elementwise tensor product Rs function, I find the usage case very helpful in understanding its purpose.

And it is a change of basis in the representation right? i.e. from a representation of multiple spherical harmonics of order Rs_1 and Rs_2 to Rs_total (irreducible)

blondegeek: It's a tensor product of two (independent) representation spaces into one. I think basis typically means you are staying "within the same space".

mariogeiger: I'm pretty sure it is a change of basis. `R1 \otimes R2` is a **reducible** representation and with this particular change of basis you make it block diagonal.

blondegeek: I definitely agree that it is the conversion of a [reducible representation to a irreducible representation](https://en.wikipedia.org/wiki/Clebsch%E2%80%93Gordan_coefficients#Tensor_product_space). I also understand why one could call it a change of basis -- we are just projecting two "kets" and onto one "bra". But, I feel that it might be confusing for other folks as to what it means to do the tensor product.

blondegeek: I think I would feel best if the matrix was explicitly called `clebsch_gordan_tensor` or something along those lines. That feels most specific and google-able for the general user.

mariogeiger: I renamed it `clebsch_gordan_tensor`

PR #22, opened 2020-02-19
functions to create helper matrices to use on tensors with only irrep…

blondegeek:
… or multiplicity indices rather than full representations -- such as spherical harmonics (irreps) and radial functions (multiplicities)

functions modified include:
* `simplifyRs` -- I factored out some code from `simplifyRs` to `conventionRs` because I needed a function that would convert Rs with just (mult, L) to (mult, L, parity) without simplifying -- didn't see another function to do that but I might've missed it.

functions added include:
* `conventionRs` -- This is `simplifyRs` as previously written but without the simplifying part. `simplifyRs` now called `conventionRs`.
* `irrep_mixing_matrixRs` -- This creates the matrix needed to map irrep channels to full representation channels. If you sort and simplify Rs before calling this matrix, it will only contain a single copy of each irrep present in Rs which is what we need for the kernel.
* `multiplicity_mixing_matrixRs` -- This creates the matrix needed to map multiplicity channels to full representation (Rs) channels. This should ONLY be used on scalar features -- such as radial functions -- otherwise it is not equivariant.


mariogeiger:
Can you add an example of usage in the docstring please?
It would be good if we can come up with more meaningful names for these functions

blondegeek:
How about for names things like `map_irreps_to_Rs` and `map_multiplicity_to_Rs`? I'll work on the doc strings.

blondegeek:
Thanks for adding the `dimRs`. Would you prefer `conventionRs` or `regularRs` for the function that regularizes the Rs to the convention form with parity?

mariogeiger:
`conventionRs` is good for me

PR #23, opened 2020-02-20
simplified kernel (something about normalization is wrong)

blondegeek:
This pull request depends on [pull request 22](https://github.com/e3nn/e3nn/pull/22).

This pull request is more for discussion than to actually be incorporated into the repository. It is a work in progress to try and simplify the kernel down to the most compact tensor contractions.

To dos:
1. Check that output matches current Kernel with no normalization.
2. Add normalization to modified Kernel using filter path info.


mariogeiger:
since PR #22 has been merged into master you should merge the current master with your branch

e3nn/kernel_mod.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Create the radial model: R+ -> R^n_path
+        # It contains the learned parameters
+        self.R = RadialModel(self.n_path)
+        self.set_of_l_filters = [L for mul, L, p in Rs_filter_simplify]
+        # Check l_filters and rep_mix have same dim
+        assert sum([2 * L + 1 for L in self.set_of_l_filters]) == irrep_mix.shape[-1]
+
+        # Register mixing matrix buffers
+        self.register_buffer('filter_clebsch_gordan', filter_mapping_matrix)
```
bkmi: maybe these names could be consistent, i.e. both filter_clebsch_gordan or both filter_mapping_matrix

bkmi: Same thing to the names below?

blondegeek: done!

e3nn/kernel_mod.py
```diff
@@ -0,0 +1,118 @@
...
+            if normalization == 'norm':
+                lm_norm = math.sqrt(2 * l_in + 1) * math.sqrt(4 * math.pi)
+            elif normalization == 'component':
+                lm_norm = math.sqrt(2 * l_out + 1) * math.sqrt(4 * math.pi)
+            return lm_norm
+
+        # Clebsch-Gordan for filter, input, output
+        # Rs_filter contains all degrees of freedom and L's for filters
+        # paths contains [l_in, l_out, l_filter for each filter channel]
+        Rs_filter, filter_clebsch_gordan, paths = SO3.tensor_productRs(Rs_in,
```
mariogeiger: @blondegeek Can you prove/argue why this is the correct function to use? (Since we are not doing the tensor product of `Rs_in` with `Rs_out`)

blondegeek: Am I correct in interpreting your question to be, why should we include the Clebsch-Gordan coefficients in `Kernel` when they are really needed for the `Convolution`? If that is so, I completely agree. And it's something we should consider perhaps storing in `Kernel` (since it's created when we create `Rs_filter`), but not applying until the `Convolution`.

Since, it is `Convolution` rather than `Kernel` itself that computes the tensor product of `Rs_in` with `Rs_filter` to get `Rs_out`, we do not need Clebsch-Gordan coefficients in `Kernel` forward.

In such a case, the forward einsum would instead read as
```
        kernel = torch.einsum('zk,zk->zk', R, Y)
        return kernel.view(*size, kernel.shape[1], kernel.shape[2])
```

blondegeek: Or perhaps, you question is instead -- how do we know that the `tensor_productRs` is giving us the same paths as the loops? If that is the question -- I'm not sure how to prove that without just printing all the paths that the loop version uses and comparing it to the `paths` list that `tensor_productRs` can now output.

Let me know which you mean and I'm happy to explain my thought process further.

mariogeiger: I realized that my question was not stated clearly enough. I appreciate the effort to guess it’s meaning. It was your second guess. 
Are the path the same? I have the impression that it is the case. 
Is the output matrix the one we really want?

mariogeiger: My question in pseudocode
```python
# (i) normal usage of tensor_productRs
Rs_out, Q = tensor_productRs(Rs1, Rs2)
out = einsum('kij,i,j->k', in1, in2)

# (ii) usage in Kernel
R_filters, Q = tensor_productRs(Rs_in, Rs_out)
out = einsum('kij,i,k->j', in, filters)
```

- (i) multiply in1 and in2 in all possible ways
- (ii) get all possible path to obtain out starting from in

These two usage are probably valid but it's non trivial for me.

mariogeiger: Then the next question will concern the normalization in the [line](https://github.com/e3nn/e3nn/blob/master/e3nn/SO3.py#L502)
```python
C = clebsch_gordan(l, l_1, l_2, cached=True) * (2 * l + 1) ** 0.5
```
it should probably depend on the usage (i) or (ii) of the function. (But I'm not sure!)

blondegeek:
I've added the normalization. I checked that the unique normalization coefficients are the same -- see `kernel_test.py`, but I think something is still off. I'm able to train with the simplified kernel, but it's definitely not as stable as the original, so it seems like I still may not have added the normalization correctly. Let me know if you have any ideas for testing.

e3nn/rs.py
```diff
@@ -203,6 +203,35 @@ def format_Rs(Rs):
...
+                num_summed_list.append(num)
+            num, cur = 0, one
+        num += 1
+    if cur is not None:
+        num_summed_list.append(num)
+    return num_summed_list
+
+
+def map_tuple_to_Rs(Rs):
+    pass
```
mariogeiger:
```suggestion
```

e3nn/rs.py
```diff
@@ -203,6 +203,35 @@ def format_Rs(Rs):
...
+    mapping_matrix = torch.zeros(dim(Rs), len(Rs))
+    start_tuple = 0
+    start_rep = 0
+    for i, (mult, L, _) in enumerate(Rs):
+        for _ in range(mult):
+            tuple_index = i 
+            rep_slice = slice(start_rep, start_rep + 2 * L + 1)
+            mapping_matrix[rep_slice, tuple_index] = 1.
+            start_rep += 2 * L + 1
+    return mapping_matrix  # [dim(Rs), irrep_dim(Rs)]
```
mariogeiger:
```suggestion
    return mapping_matrix  # [dim(Rs), len(Rs)]
```

e3nn/rs.py
```diff
@@ -203,6 +203,35 @@ def format_Rs(Rs):
...
+
+def map_tuple_to_Rs(Rs):
+    pass
+    Rs = convention(Rs)
+    mapping_matrix = torch.zeros(dim(Rs), len(Rs))
+    start_tuple = 0
+    start_rep = 0
+    for i, (mult, L, _) in enumerate(Rs):
+        for _ in range(mult):
+            tuple_index = i
```
mariogeiger:
```suggestion
    for tuple_index, (mult, L, _) in enumerate(Rs):
        for _ in range(mult):
```

blondegeek:
Fixed and merged with https://github.com/e3nn/e3nn/pull/50

Issue #13, opened 2020-02-12
get continious integration working


bkmi:

bkmi:
solved with #27

PR #24, opened 2020-02-26
Added pytest to continious integration

bkmi:
Solves issue #13 

I deleted the all.py since it is no longer necessary. Simply run
```shell
pytest
```
from the main directory.


mariogeiger:
Cool, let see if it works !^^ 
Because I had the problem that pytest was lauching tests in lie_learn and they failed

mariogeiger:
you have an indentation error

mariogeiger:
For this purpose I think you should directly push on master without passing by PR otherwise you will have to wait for me too long

bkmi:
I was way too fast, there are other problems I wasn't expecting. I suggest reverting the pull request for now (although it does only affect the tests, not the actual code)

mariogeiger:
ok

bkmi:
I am not having the issues with testing lie_learn like you experienced. I found a bit of a clue which I posted on [the issue there](https://github.com/AMLab-Amsterdam/lie_learn/issues/11).

PR #27, opened 2020-02-26
CI working

bkmi:
Again, actually fixed issue #13. some tests are ignored.
lie_learn is no longer tested as it is not installed in develop mode. 
periodic convolution is also not tested since it involves adding pymatgen to requirements. Not sure if we want this.


mariogeiger:
Thanks a lot!

PR #29, opened 2020-02-28
Added KernelConv, reintroduced kernel without KernelFn, reintroduced ConvolutionFn

bkmi:
This introduces a new feature KernelConv.

Adds custom forward and backward function for convolution.

Adds automatic backwards (normal pytorch behavior) Kernel. I suggest we use this by default. Right now it is listed as a separate function which has a corresponding test case.

Although it is only for a QM9 style problem (i.e. shape of input) KernelConv combines the two operations just as was requested in the optimization project.

### DIFF
--- a/e3nn/kernel.py
+++ b/e3nn/kernel.py
@@ -8,7 +8,7 @@ from e3nn import o3, rs
 
 class Kernel(torch.nn.Module):
     def __init__(self, Rs_in, Rs_out, RadialModel, get_l_filters=o3.selection_rule, sh=o3.spherical_harmonics_xyz, normalization='norm'):
-        '''
+        """
         :param Rs_in: list of triplet (multiplicity, representation order, parity)
         :param Rs_out: list of triplet (multiplicity, representation order, parity)
         :param RadialModel: Class(d), trainable model: R -> R^d
@@ -17,7 +17,7 @@ class Kernel(torch.nn.Module):
         :param normalization: either 'norm' or 'component'
         representation order = nonnegative integer
         parity = 0 (no parity), 1 (even), -1 (odd)
-        '''
+        """
         super().__init__()
 
         self.Rs_in = rs.simplify(Rs_in)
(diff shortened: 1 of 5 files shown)


bkmi:
Just realized that I don't think this is compatible with GatedBlock. Maybe hold off on accepting the changes until it does?

e3nn/kernel.py
```diff
@@ -249,3 +249,73 @@ def backward(ctx, grad_kernel):
 
         del ctx
         return grad_Y, grad_R, None, None, None, None, None
+
+
+class KernelAutoBackward(Kernel):
+    def __init__(self, Rs_in, Rs_out, RadialModel, get_l_filters=o3.selection_rule, sh=o3.spherical_harmonics_xyz, normalization='norm'):
+        super(KernelAutoBackward, self).__init__(Rs_in, Rs_out, RadialModel, get_l_filters, sh, normalization)
+
+    def forward(self, r):
```
mariogeiger: What is the difference with the `forward` of `KernelFn` ?

e3nn/point/kernelconv.py
```diff
@@ -0,0 +1,103 @@
...
+        norm_coef = getattr(self, 'norm_coef')
+        norm_coef = norm_coef[:, :, (radii == 0).type(torch.long)]  # [l_out, l_in, batch, a, b]
+
+        kernel_conv = kernel_conv_automatic_backward(
+            features, y, r, norm_coef, self.Rs_in, self.Rs_out, self.get_l_filters, self.set_of_l_filters
+        )
+        return kernel_conv * mask.unsqueeze(-1)
+
+
+def kernel_conv_automatic_backward(features, Y, R, norm_coef, Rs_in, Rs_out, get_l_filters, set_of_l_filters):
```
mariogeiger: the name is super super weird. why is is called `automatic_backward` ? I don't want to call al my function `automatic_backward` just because I don't provide the custom backward.

e3nn/kernel.py
```diff
@@ -249,3 +249,73 @@ def backward(ctx, grad_kernel):
 
         del ctx
         return grad_Y, grad_R, None, None, None, None, None
+
+
+class KernelAutoBackward(Kernel):
```
mariogeiger: I don't like the name of this class

bkmi:
oops! This was a mistake!! I'm sorry about that!

bkmi:
I did a forced reset to unto these changes. I have the changes in my branch and will attempt another pull request when it is in better shape.

I think everything should be okay now, sorry about that.

mariogeiger:
No worries

Issue #18, opened 2020-02-12
Merge kernel and convolution


blondegeek:
This will help alleviate memory bottlenecks for training larger models.

PR #30, opened 2020-03-03
Introduced KernelConv, Introduced custom backward pass options for kernel and convolution and KernelConv

bkmi:
I think @mariogeiger might not like that there is a keyword argument. The biggest offender is probably the new convolution. Let me know if you think I should change it.

Another option is to instead create an attribute of the object which says whether or not to use custom_backward or not. I didn't like this as much since it is less flexible; however, it is more "object oriented."

The introduction of KernelConv solves issue #18 

--

The default behavior is to use custom backward for KernelConv and the automatic backward for kernel and convolution.

When using KernelConv with custom_backward=True in qm9 I was able to do 1000 molecule epochs in 11 seconds. When using custom_backward=False in qm9 the same number of molecules were iterated over in nearly 18 seconds.

It seems the behavior of these functions depends on the context. I believe it calls for further looking into whether or not custom_backward for Kernel should be enabled by default. (In this pull request custom_backward = False is default for Kernel and Convolution)

### DIFF
--- a/e3nn/kernel.py
+++ b/e3nn/kernel.py
@@ -7,7 +7,8 @@ from e3nn import o3, rs
 
 
 class Kernel(torch.nn.Module):
-    def __init__(self, Rs_in, Rs_out, RadialModel, get_l_filters=o3.selection_rule, sh=o3.spherical_harmonics_xyz, normalization='norm'):
+    def __init__(self, Rs_in, Rs_out, RadialModel, get_l_filters=o3.selection_rule, sh=o3.spherical_harmonics_xyz,
+                 normalization='norm', custom_backward=False):
         """
         :param Rs_in: list of triplet (multiplicity, representation order, parity)
         :param Rs_out: list of triplet (multiplicity, representation order, parity)
@@ -15,10 +16,12 @@ class Kernel(torch.nn.Module):
         :param get_l_filters: function of signature (l_in, l_out) -> [l_filter]
         :param sh: spherical harmonics function of signature ([l_filter], xyz[..., 3]) -> Y[m, ...]
         :param normalization: either 'norm' or 'component'
+        :param custom_backward: call KernelFn rather than using automatic differentiation
         representation order = nonnegative integer
         parity = 0 (no parity), 1 (even), -1 (odd)
         """
         super().__init__()
+        self.custom_backward = custom_backward
 
         self.Rs_in = rs.simplify(Rs_in)
         self.Rs_out = rs.simplify(Rs_out)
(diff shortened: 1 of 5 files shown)


e3nn/kernel.py
```diff
@@ -125,19 +125,71 @@ def forward(self, r):
...
 class KernelFn(torch.autograd.Function):
     @staticmethod
     def forward(ctx, Y, R, norm_coef, Rs_in, Rs_out, get_l_filters, set_of_l_filters):
-        """
-        :param Y: tensor [l_filter * m_filter, batch]
-        :param R: tensor [batch, l_out * l_in * mul_out * mul_in * l_filter]
-        :param norm_coef: tensor [l_out, l_in, batch]
-        :return: tensor [batch, l_out * mul_out * m_out, l_in * mul_in * m_in]
-        """
+        f"""{kernel_fn_forward.__doc__}"""
```
mariogeiger: fancy

e3nn/kernel.py
```diff
@@ -153,51 +205,7 @@ def forward(ctx, Y, R, norm_coef, Rs_in, Rs_out, get_l_filters, set_of_l_filters
...
-                    C = o3.clebsch_gordan(l_out, l_in, l_filter, cached=True, like=kernel)  # [m_out, m_in, m]
-
-                    # note: The multiplication with `sub_R` could also be done outside of the for loop
-                    K += torch.einsum("ijk,kz,zuv,z->zuivj", (C, sub_Y, sub_R[..., k], sub_norm_coef))  # [batch, mul_out, m_out, mul_in, m_in]
-
-                if K is not 0:
-                    kernel[:, s_out, s_in] = K.contiguous().view_as(kernel[:, s_out, s_in])
-
-        return kernel
+        return kernel_fn_forward(Y, R, norm_coef, ctx.Rs_in, ctx.Rs_out, ctx.get_l_filters, ctx.set_of_l_filters)
```
mariogeiger: cool that you reuse the same code

tests/point/kernelconv_test.py
```diff
@@ -0,0 +1,145 @@
...
+        for mul_out, l_out, p_out in KC.Rs_out:
+            for mul_in, l_in, p_in in KC.Rs_in:
+                l_filters = KC.get_l_filters(l_in, p_in, l_out, p_out)
+                n_path += mul_out * mul_in * len(l_filters)
+
+        batch = 2
+        atoms = 3
+
+        true_false = [False, True]
+        options = set(product(true_false, repeat=3)) - {(False, False, False)}
```
mariogeiger: I didn't knew that the minus operator was working for sets

bkmi:
> Very nice code

Thanks for the compliment!! :)

PR #32, opened 2020-03-12
add property to keep track of which spherical harmonic is attached to…

blondegeek:
… which radial function

### DIFF
--- a/e3nn/kernel.py
+++ b/e3nn/kernel.py
@@ -49,6 +49,7 @@ class Kernel(torch.nn.Module):
 
         n_path = 0
         set_of_l_filters = set()
+        list_of_l_filters = []
 
         for i, (mul_out, l_out, p_out) in enumerate(self.Rs_out):
             # consider that we sum a bunch of [lambda_(m_out)] vectors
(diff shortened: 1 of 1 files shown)


e3nn/kernel.py
```diff
@@ -57,6 +58,8 @@ def lm_normalization(l_out, l_in):
             for mul_in, l_in, p_in in self.Rs_in:
                 l_filters = self.get_l_filters(l_in, p_in, l_out, p_out)
                 num_summed_elements += mul_in * len(l_filters)
+                list_of_l_filters += [l for l in l_filters for i in range
```
mariogeiger: don't you want to save more information? like a tuple `(l_in, p_in, l_out, p_out, l)`?

blondegeek:
My motivation for adding this list is to facilitate plotting kernels. I'm happy for the list to store the full information, but I only need l_filter. See [this in progress function](https://github.com/blondegeek/e3nn_tutorial/blob/volume/spherical.py#L207) (that I hope to add as a method to Kernel) that I'm using to visualize the kernel and [this notebook](https://github.com/blondegeek/e3nn_tutorial/blob/volume/plot_filter.ipynb) for an example of how to use it.

e3nn/kernel.py
blondegeek: My motivation for adding this list is to facilitate plotting kernels. I'm happy for the list to store the full information, but I only need l_filter. See [this in progress function](https://github.com/blondegeek/e3nn_tutorial/blob/volume/spherical.py#L207) (that I hope to add as a method to Kernel) that I'm using to visualize the kernel and [this notebook](https://github.com/blondegeek/e3nn_tutorial/blob/volume/plot_filter.ipynb) for an example of how to use it.

mariogeiger: As you wish, I merge!

PR #36, opened 2020-03-19
added a simple default convolutional network with gated activations

blondegeek:
I found myself recreating this network over and over again, so I though it might be a nice default to have. Please feel free to modify, such moving activations and radial models as keywords.


mariogeiger:
Nice

JoshRackers:
Highly support this!

PR #38, opened 2020-03-21
clean code of GatedConvNetwork

mariogeiger:


mariogeiger:
I also put `swish` instead of `ShiftedSoftplus` apparently it's a trendy activation function ^^
And it has the property that `phi(0)=0`

PR #39, opened 2020-03-21
Change the order of indices in Batchnorm

mariogeiger:
- changed the order of indices `[batch, features, spacial]` -> `[batch, spacial, features]`
- complete docstring

### DIFF
--- a/e3nn/batchnorm.py
+++ b/e3nn/batchnorm.py
@@ -1,4 +1,4 @@
-# pylint: disable=C,R,E1101
+# pylint: disable=no-member, arguments-differ, missing-docstring, invalid-name
 import torch
 import torch.nn as nn
 
@@ -6,7 +6,18 @@ import torch.nn as nn
 class BatchNorm(nn.Module):
     def __init__(self, Rs, eps=1e-5, momentum=0.1, affine=True, reduce='mean'):
         '''
+        Batch normalization layer for orthonormal representations
+        It normalizes by the norm of the representations.
+        Not that the norm is invariant only for orthonormal representations.
+        Irreducible representations `o3.irr_repr` are orthonormal.
+
+        input shape : [batch, [spacial dimensions], stacked orthonormal representations]
+
         :param Rs: list of tuple (multiplicity, dimension)
+        :param eps: avoid division by zero when we normalize by the variance
+        :param momentum: momentum of the running average
+        :param affine: do we have weight and bias parameters
+        :param reduce: method to contract over the spacial dimensions
         '''
         super().__init__()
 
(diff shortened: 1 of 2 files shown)


tests/batchnorm_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Tests(unittest.TestCase):
+    def test_that_it_runs(self):
+        Rs = [(3, 1), (4, 3)]
+        m = BatchNorm(Rs)
+
+        m.to(dtype=torch.float64)
+        x = torch.randn(10, sum(mul * d for mul, d in Rs), dtype=torch.float64)
+        m(x)
```
blondegeek: Should we be testing any properties of the output?

tests/batchnorm_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+        m.to(dtype=torch.float32)
```
blondegeek: Rather than just two examples, do we want to add 10 for test?
Additionally, should we test that on average, for inputs of varying variance (perhaps an additional outside look to the number of tests), that batch norm is able to bring the variance back to 1?

tests/batchnorm_test.py
mariogeiger: Yes we should do more than that :+1:

tests/batchnorm_test.py
mariogeiger: You are right, it was laziness

tests/batchnorm_test.py
```diff
@@ -0,0 +1,42 @@
...
+        m = BatchNorm(Rs, normalization='component')
+
+        x = torch.randn(batch, n, sum(mul * d for mul, d in Rs)).mul(5.0).add(10.0)
+        x = m(x)
+
+        a = x[..., :3]  # [batch, space, mul]
+        assert a.mean([0, 1]).abs().max() < 1e-10
+        assert a.pow(2).mean([0, 1]).sub(1).abs().max() < 1e-5
+
+        a = x[..., 3:].reshape(batch, n, 4, 3)  # [batch, space, mul, repr]
```
blondegeek: Super cool. Thanks for the tests. Should there also be checks on variance or is mean sufficient?

e3nn/batchnorm.py
```diff
@@ -28,8 +39,14 @@ def __init__(self, Rs, eps=1e-5, momentum=0.1, affine=True, reduce='mean'):
             self.register_parameter('weight', None)
             self.register_parameter('bias', None)
 
+        assert isinstance(reduce, str), "reduce should be passed as a string value"
+        assert reduce in ['mean', 'max'], "reduce needs to be 'mean' or 'max'"
         self.reduce = reduce
 
+        assert isinstance(normalization, str), "normalization should be passed as a string value"
+        assert normalization in ['norm', 'component'], "normalization needs to be 'norm' or 'component'"
```
blondegeek: Ah -- nice feature!

PR #41, opened 2020-03-29
implement Linear & add equivariance tests

mariogeiger:


tests/linear_test.py
```diff
@@ -0,0 +1,39 @@
...
+from e3nn import rs
+
+
+class Tests(unittest.TestCase):
+    def test1(self):
+        torch.set_default_dtype(torch.float64)
+        Rs_in = [(50, 0), (150, 1), (50, 0), (100, 2)]
+        Rs_out = [(2, 0), (1, 1), (1, 2), (3, 0)]
+
+        torch.manual_seed(0)
```
blondegeek: This seems like an awkward test since it depends on a manual_seed producing a specific value.

I ran the following locally,
```
lin_vals = []
lin_mod_vals = []
for i in range(1000):
    Rs_in = [(5, 0), (3, 1), (2, 0), (7, 2)] 
    Rs_out = [(5, 0), (1, 1), (1, 2), (3, 0)] 
    torch.manual_seed(i)
    linmod = LinearMod(Rs_in, Rs_out)
    lin = Linear(Rs_in, Rs_out)
    N = 2
    input = torch.randn(1, N, rs.dim(Rs_in))
    lin_vals.append(lin(input).pow(2).mean(0).sub(1).abs().max().detach().numpy())
    lin_mod_vals.append(linmod(input).pow(2).mean(0).sub(1).abs().max().detach().numpy())
```
And got the following plots for Linear and my modified Linear code, respectively...
image
image

Is this what you would expect?

mariogeiger: What we want to test is that the output is gaussian.

```python
torch.set_default_dtype(torch.float64)
Rs_in = [(50, 0), (30, 1), (20, 0), (70, 2)] 
Rs_out = [(2, 0), (1, 1), (1, 2), (3, 0)]

lin = Linear(Rs_in, Rs_out)
features = torch.randn(10000, rs.dim(Rs_in))
features = lin(features)

plt.hist(features.detach().flatten(), density=True, bins=100)
x = torch.linspace(-3, 3, 100)
plt.plot(x, x.pow(2).div(-2).exp().div(math.sqrt(2 * math.pi)))
```
image

tests/linear_test.py
```diff
@@ -0,0 +1,51 @@
...
+        p = torch.histc(features, bins, left, right) / features.numel() / bin_width
+        q = x.pow(2).div(-2).exp().div(math.sqrt(2 * math.pi))  # Normal law
+
+        # import matplotlib.pyplot as plt
+        # plt.plot(x, p)
+        # plt.plot(x, q)
+        # plt.show()
+
+        Dkl = ((p + 1e-100) / q).log().mul(p).sum()  # Kullback-Leibler divergence of P || Q
+        self.assertLess(Dkl, 0.1)
```
mariogeiger: Here is the new test with Kullback-Leibler divergence

PR #42, opened 2020-03-30
simplified linear

blondegeek:
@mariogeiger I've rewritten Linear using rs.tensor_product and partial(o3.selection_rule, lmax=0). I still don't have the variance matching so I suspect whatever normalization goof I've made for the simplified kernel, I'm making a similar one for the Linear layer. I'll keep working on it, but take a look when you have a chance.

### DIFF
--- a/e3nn/linear_mod.py
+++ b/e3nn/linear_mod.py
@@ -27,11 +27,15 @@ class Linear(torch.nn.Module):
         )
 
         num_summed_list = rs.num_summed_elements(paths)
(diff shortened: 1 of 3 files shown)


blondegeek:
@mariogeiger Can you check the variance on this layer as per my comment:
https://github.com/e3nn/e3nn/pull/41/files#r399917735

I'm now not sure whether or not there is a variance / normalization issue. I likely use the weights in a different order than you, so setting the weights equal will not result in equal output.

mariogeiger:
Done in #52

PR #46, opened 2020-03-30
Add feature interaction to GatedConvNetwork (needs to be tested)

blondegeek:


e3nn/networks.py
```diff
@@ -27,9 +31,16 @@ def __init__(self, Rs_in, Rs_hidden, Rs_out, lmax, layers=3, max_radius=1.0, num
 
         K = partial(Kernel, RadialModel=RadialModel, get_l_filters=partial(o3.selection_rule, lmax=lmax))
```
mariogeiger:
```suggestion
        K = partial(Kernel, RadialModel=RadialModel, get_l_filters=partial(o3.selection_rule_in_out_sh, lmax=lmax))
```
Now you need to use this function for the selection of the SH that satisfies the parity

e3nn/networks.py
```diff
@@ -27,9 +31,16 @@ def __init__(self, Rs_in, Rs_hidden, Rs_out, lmax, layers=3, max_radius=1.0, num
...
 
-        def make_layer(Rs_in, Rs_out):
+        def make_layer(Rs_in, Rs_out, feature_product=feature_product):
```
mariogeiger:
```suggestion
        def make_layer(Rs_in, Rs_out):
```
no need to pass this argument

mariogeiger:
I modified two small things

PR #50, opened 2020-03-31
tensor_product(input1, input2, output)

mariogeiger:
This pull request is on top of #45

### DIFF
--- a/e3nn/rs.py
+++ b/e3nn/rs.py
@@ -24,7 +24,7 @@ def rep(Rs, alpha, beta, gamma, parity=None):
         return o3.direct_sum(*[(p ** parity) * o3.irr_repr(l, *abc) for mul, l, p in simplify(Rs) for _ in range(mul)])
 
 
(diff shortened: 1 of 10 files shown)


e3nn/rs.py
```diff
@@ -271,14 +271,45 @@ def sorted_truncated_tensor_product(Rs_1, Rs_2, lmax):
...
+    if isinstance(input1, list) and isinstance(input2, list):
+        return _tensor_product_in_in(input1, input2, output, normalization)
+
+    if isinstance(input1, list) and isinstance(output, list):
+        return _tensor_product_in_out(input1, input2, output, normalization)
+
+    if isinstance(input2, list) and isinstance(output, list):
+        Rs_in1, Q = _tensor_product_in_out(input2, input1, output, normalization)
+        Q = torch.einsum('kij->kji', Q)
+        return Rs_in1, Q
```
mariogeiger: Sofar it's just like that but later I will look if the functions can be really merged into one

PR #55, opened 2020-04-02
KernelLinear

mariogeiger:
1. Introduce (non _mod version of) `KernelLinear` 
2. Use it in `Kernel`

2 imply that `norm_coef` is no longer shape 
```
[Rs_out, Rs_in, 2]
```

but 
```
[Rs_out, Rs_in]
```

@bkmi I had to adapt consistently `KernelConv` and I don't like that I had to assume that index `a` is of the same nature as index `b`. Please have a look at my modifications and be aware that I might introduces bugs in the case of `{a} != {b}`.


bkmi:
I'm not bothered by having B != A. It's better design.

e3nn/networks.py
```diff
@@ -55,7 +55,7 @@ def forward(self, input, geometry):
 
         if self.feature_product:
             for tp, lin, conv, act in self.layers[:-1]:
-                output = tp(output)
+                output = tp(output, output)
```
blondegeek, line 58: Oops! Thanks for fixing that!

PR #57, opened 2020-04-03
Porting SphericalTensor over to e3nn.tensor (in progress)

blondegeek:


blondegeek:
I suspect that the current e3nn.tensor and e3nn.util.plot will need to be modified together, especially for all the cached spherical grids for quickly generating spherical harmonic plots. There is also some likely redundancy with some of the math operations (direct sums, tensor products, ...) that should be rewritten with native e3nn functions.

mariogeiger:
Please request a review when you fixed the redundancy.
You can change `e3nn.util.plot` like you want, you can even remove it.

blondegeek:
Superseded by https://github.com/e3nn/e3nn/pull/71

PR #58, opened 2020-04-03
Added LearnableMultiplicityBispectrum

blondegeek:
LearnableBispectrum get's big really fast, LearnableMultiplicityBispectrum mitigates some of this. We might be able to merge them into the same object but they're pretty short and it might not hurt to have a more general and less general class.


e3nn/tensor_product.py
```diff
@@ -109,3 +109,21 @@ def forward(self, input):
...
+        self.tp = LearnableTensorProduct(single_mul_Rs_in, single_mul_Rs_in, mul_in, Rs_hidden,
+                                         partial(o3.selection_rule, lmax=self.lmax))
+        # Dot product
+        self.dot = LearnableTensorProduct(Rs_hidden, single_mul_Rs_in * mul_in, 1, 
+                                          [(mul_out, 0)], partial(o3.selection_rule, lmax=0))
+
+    def forward(self, input):
+        output = input
+        output = self.tp(output, output)
+        return self.dot(output, input)
```
mariogeiger, line 129:
```suggestion
        return self.dot(self.tp(input, input), input)
```

e3nn/tensor_product.py
```diff
@@ -109,3 +109,21 @@ def forward(self, input):
         output = input
         output = self.tp(output, output)
         return self.dot(output, input)
+
+
+class LearnableMultiplicityBispectrum(torch.nn.Module):
+    def __init__(self, mul_in, single_mul_Rs_in, mul_hidden, mul_out):
+        super().__init__()
+        self.lmax = max(l for mul, l in single_mul_Rs_in)
+        Rs_hidden = [(mul_hidden, l) for l in range(self.lmax + 1)]
```
mariogeiger, line 118: Can you make it work for parity?

e3nn/tensor_product.py
```diff
@@ -109,3 +109,21 @@ def forward(self, input):
         output = input
         output = self.tp(output, output)
         return self.dot(output, input)
+
+
+class LearnableMultiplicityBispectrum(torch.nn.Module):
+    def __init__(self, mul_in, single_mul_Rs_in, mul_hidden, mul_out):
```
mariogeiger, line 115: Why not ask for `Rs_hidden` in argument? Do we really need to impose the kind of representation?

mariogeiger:
While looking at the code I saw that the order of the argument of `LearnableTensorProduct` are not very nice.

```python
class LearnableTensorProduct(torch.nn.Module):
    def __init__(self, Rs_mid_1, Rs_mid_2, mul_mid, Rs_out, selection_rule=o3.selection_rule):
```

I propose to rename `mid -> in` and `mul_mid -> groups`
```python
class LearnableTensorProduct(torch.nn.Module):
    def __init__(self, Rs_in1, Rs_in2, selection_rule, Rs_out, groups=1):
```

e3nn/tensor_product.py
blondegeek: Sure thing!

e3nn/tensor_product.py
blondegeek: 👍

e3nn/tensor_product.py
blondegeek: That's true. The same could be said for `LearnableBispectrum`. I'll modify for both.

blondegeek:
Changes made but needs tests. Please feel free to review the code itself, but please do not merge yet.

e3nn/tensor_product.py
```diff
@@ -73,39 +73,70 @@ def forward(self, features_1, features_2):
 
 
 class LearnableTensorProduct(torch.nn.Module):
-    def __init__(self, Rs_mid_1, Rs_mid_2, mul_mid, Rs_out, selection_rule=o3.selection_rule):
+    def __init__(self, Rs_mid_1, Rs_mid_2, Rs_out,
```
mariogeiger, line 76: What about `in`? `mid` means middle?
```suggestion
    def __init__(self, Rs_in1, Rs_in2, Rs_out,
```

e3nn/tensor_product.py
```diff
@@ -73,39 +73,70 @@ def forward(self, features_1, features_2):
...
+                 selection_rule=o3.selection_rule, groups=1):
```
mariogeiger, line 77: Also I propose that the order reflects the input to output order therefore I would put `selection_rule` before `Rs_out`

e3nn/tensor_product.py
```diff
@@ -73,39 +73,70 @@ def forward(self, features_1, features_2):
...
+        if lmax is None:
+            self.lmax = max(l for mul, l, _ in self.Rs_hidden)
+        else:
+            self.lmax = lmax
         # Learnable tensor product of signal with itself
-        self.tp = LearnableTensorProduct(Rs_in, Rs_in, 1, Rs_hidden,
-                                         partial(o3.selection_rule, lmax=self.lmax))
+        self.tp = LearnableTensorProduct(
+            self.Rs_in, self.Rs_in, self.Rs_hidden,
+            partial(o3.selection_rule, lmax=self.lmax), groups=1)
```
mariogeiger, line 109: The selection of the `l` that are present in the output of the `LearnableTensorProduct` should be done automatically inside `LearnableTensorProduct`. Because never you will allow a l that is not in the output rep. It will actually raise an error in `Linear`.

e3nn/tensor_product.py
```diff
@@ -73,39 +73,70 @@ def forward(self, features_1, features_2):
...
         # Dot product
-        self.dot = LearnableTensorProduct(Rs_hidden, Rs_in, 1,
-                                          [(mul_out, 0)], partial(o3.selection_rule, lmax=0))
+        self.dot = LearnableTensorProduct(
+            self.Rs_hidden, self.Rs_in, [(mul_out, 0)],
+            partial(o3.selection_rule, lmax=0), groups=1)
```
mariogeiger, line 113: Same here, you should not need to specify the selection rule

e3nn/tensor_product.py
```diff
@@ -73,39 +73,70 @@ def forward(self, features_1, features_2):
...
 
     def forward(self, input):
-        output = input
-        output = self.tp(output, output)
-        return self.dot(output, input)
+        return self.dot(self.tp(input, input), input)
```
mariogeiger, line 116: Since you do `tp(x, x)` I will implement a `tensor_square` operation that does not compute the odd part that is zero

e3nn/tensor_product.py
```diff
@@ -73,39 +73,70 @@ def forward(self, features_1, features_2):
...
-        x2 = x2.view(*size, self.mul_mid, -1)
+        x1 = x1.view(*size, self.groups, -1)
+        x2 = x2.view(*size, self.groups, -1)
 
         x = self.tp(x1, x2).view(*size, -1)
         x = self.lin(x)
         return x
 
 
 class LearnableBispectrum(torch.nn.Module):
```
mariogeiger, line 97: Just wondering if we can find a more "down to earth" that says just what it does and not how it comes from because I have the impression that at the end it's simpler that where it comes from

blondegeek:
Closing this for now. Will open a new pull request if needed.

PR #60, opened 2020-04-03
improve S2 non linearity with SOFT grid

mariogeiger:
Part of the merge of s2cnn into e3nn.

Still have to merge so3 non linearity


mariogeiger:
In this nonlinearity the equivariance is not exact.
I'm looking at the error of equivariance wrt to the resolution of the grid used to map the signal.

image

- It's globally quite low (to be compared to 1).
- For swish it even goes to the machine precision
- For swish it goes up again, I don't understand that part !

mariogeiger:
**Bug fixed !**
image

PR #63, opened 2020-04-05
image

mariogeiger:


e3nn/image/convolution.py
```diff
@@ -3,22 +3,32 @@
...
+        self.kernel = FrozenKernel(Rs_in, Rs_out, R, r, normalization='component')
         self.kwargs = kwargs
 
     def forward(self, features):
-        k = torch.einsum('xyzij->ijxyz', self.kernel(self.r))
-        k.mul_(2 / k[0, 0].numel() ** 0.5)
+        """
+        :param tensor features: [batch, j, x, y, z]
+        """
+        k = torch.einsum('xyzij->ijxyz', self.kernel())
```
blondegeek, line 32: I'm worries about the speed of this transpose. Is there a way to avoid it in the forward?

e3nn/kernel_mod.py
```diff
@@ -109,3 +109,83 @@ def forward(self, r, r_eps=0, **_kwargs):
         kernel[radii <= r_eps] = kernel2
 
         return kernel.view(*size, *kernel2.shape)
+
+
+class FrozenKernel(torch.nn.Module):
```
blondegeek, line 114: This is great! Love the name!

e3nn/image/convolution.py
```diff
@@ -3,22 +3,32 @@
...
-from e3nn.non_linearities import rescaled_act
-from e3nn.kernel import Kernel
+from e3nn.non_linearities.rescaled_act import swish
+from e3nn.kernel_mod import FrozenKernel
 from e3nn.radial import CosineBasisModel
 
 
 class Convolution(torch.nn.Module):
-    def __init__(self, Rs_in, Rs_out, size, **kwargs):
+    def __init__(self, Rs_in, Rs_out, size, steps=(1, 1, 1), **kwargs):
```
blondegeek, line 12: Steps is a great way to treat anisotropic cells.

e3nn/image/convolution.py
mariogeiger: We can redefine `Kernel` but I prefer if we use this class and measure the time spent to transpose compare to the convolution. If it's less than 1% I will not touch it

e3nn/kernel_mod.py
mariogeiger: Thanks ^^

PR #64, opened 2020-04-06
Included bessel radial, math for functions, and fixed an import problem

bkmi:


mariogeiger:
please merge master into your branch, I will look at your code, I'm wondering why you need to create a hole new class for your new basis function

bkmi:
The reason is that Bessel needs to implement a piecewise if |r| > c then f(r) = 0.

I can try to figure out if I can do it with existing classes.

e3nn/radial.py
```diff
@@ -91,10 +97,40 @@ def basis(x):
...
+        self.register_buffer('n_scaled', n_scaled)
+        self.factor = math.sqrt(2/max_radius)
+        self.max_radius = max_radius
+        self.epsilon = epsilon
+
+        self.f = FC(number_of_basis, out_dim, h=h, L=L, act=act)
+
+    def basis(self, x):
+        assert x.ndim == 1
+        x_within_cutoff = torch.logical_not((x < 0.0) + (x > self.max_radius)) * x
```
mariogeiger, line 130:
```suggestion
        x_within_cutoff = ((x >= 0.0) * (x <= self.max_radius)) * x
```

mariogeiger:
> The reason is that Bessel needs to implement a piecewise if |r| > c then f(r) = 0.
> 
> I can try to figure out if I can do it with existing classes.

Yes and also because you are not just translating a function

PR #67, opened 2020-04-09
Big changes in spherical harmonics

mariogeiger:
- new file `rsh.py` for Real Spherical Harmonic
- all functions request list of L's  (to simplify the code, sorry user)
- all functions request pytorch tensor (to simplify the code, sorry user)
- home-made computation using legendre (stop using lie_learn)
- remove `backward` version but use backwadable code when input requires grad

### DIFF
--- a/e3nn/rsh.py
+++ b/e3nn/rsh.py
@@ -1,4 +1,4 @@
-# pylint: disable=not-callable, no-member, invalid-name, line-too-long, unexpected-keyword-arg, too-many-lines
+# pylint: disable=not-callable, no-member, invalid-name, line-too-long
 """
 Real Spherical Harmonics equivariant with respect to o3.rot and o3.irr_repr
 """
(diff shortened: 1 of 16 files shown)


e3nn/rsh.py
```diff
@@ -0,0 +1,167 @@
...
+def spherical_harmonics_expand_matrix(lmax):
+    """
+    :return: tensor [l, m, l * m]
+    """
+    m = torch.zeros(lmax + 1, 2 * lmax + 1, sum(2 * l + 1 for l in range(lmax + 1)))
+    i = 0
+    for l in range(lmax + 1):
+        m[l, lmax - l: lmax + l + 1, i:i + 2 * l + 1] = torch.eye(2 * l + 1)
+        i += 2 * l + 1
+    return m
```
bkmi, line 22: Should we mention where this is used so a new person would understand what it is for better? I.e. in the KernelLinear

e3nn/rsh.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if l == 0:
+        return z.new_ones(*z.size(), 1)
+    if l == 1:
+        return torch.stack([-0.5 * sqz2, z, sqz2], dim=-1)
+
+    plm = [(1 - 2 * abs(l - 2 * l // 2)) * hsqz2 ** l / fac]
+    plm.append(-plm[0] * l * ihsqz2)
+    for mr in range(1, l):
+        plm.append((mr - l) * ihsqz2 * plm[mr] - (2 * l - mr + 1) * mr * plm[mr - 1])
```
bkmi, line 49: Maybe I'm being ambitious, but should we do this in cython?

mariogeiger:
> Does this mean we are not going to use cuda real spherical harmonics anymore?

I removed it because it was ugly coded and it has to be recoded the day was are able to do an installation procedure with optional compilation

e3nn/rsh.py
mariogeiger: Ok!

e3nn/rsh.py
mariogeiger: I don't want to use cython, it will complexify the installation process again. This code is used only in the case of `requires_grad=True` so I don't care much.

bkmi:
I think it looks good. Maybe let's try to get it before we upload the first versions onto pypi.

PR #70, opened 2020-04-15
Getting ready for pypi, fixes setup_cuda.py

bkmi:
I don't know if I like the multiple setup.py though. If we decide to go with it, I believe this is a solution. I'm attempting the install now.


bkmi:
In fact, I get an error when attempting to use the `setup_cuda.py`. I don't think it is possible to use python_cuda.py it fails at gcc.

bkmi:
setup_cuda.py had a bug from origin/master.

pack.py
```diff
@@ -0,0 +1,58 @@
...
+    if version_match:
+        return version_match.group(1)
+    raise RuntimeError("Unable to find version string.")
+
+
+def setup_kwargs(**kwargs):
+    return kwargs
+
+
+KWARGS = setup_kwargs(
```
mariogeiger, line 29:
```suggestion
KWARGS = dict(
```

setup_cuda.py
```diff
@@ -1,31 +1,18 @@
...
+from setuptools import setup
 from torch.utils.cpp_extension import BuildExtension, CUDAExtension
+from pack import KWARGS
 
 ext_modules = [
     CUDAExtension('e3nn.rsh',
-                  sources=['src/rsh/rsh_bind.cpp',
-                           'src/rsh/rsh_cuda.cu'],
+                  sources=['src/real_spherical_harmonics/rsh_bind.cpp',
+                           'src/real_spherical_harmonics/rsh_cuda.cu'],
```
mariogeiger, line 9: Oops sorry

PR #78, opened 2020-04-26
add parity to SphericalTensor

mariogeiger:


e3nn/spherical_tensor.py
```diff
@@ -46,20 +46,42 @@ def adjusted_projection(vectors, lmax):
 
 
 class SphericalTensor():
-    def __init__(self, signal, mul, lmax):
+    def __init__(self, signal, mul, lmax, p_val=0, p_arg=0):
```
blondegeek, line 49: Will p_val always be an int or will it be specified per L?

mariogeiger: int. Otherwise it would lead to weird parity transformations for a signal on the sphere. 
It’s explained in the docstring.

PR #82, opened 2020-05-06
Sparse operations

blondegeek:
Since the Q matrix gets really big for `kernel_mod.Kernel`, I wanted to try sparse operations from https://github.com/rusty1s/pytorch_sparse. The first part of this is making a sparse tensor product. I think we could get the sparseness down by even more by forgoing the use of torch.eye in the inner loop but for now there's at least a halving of memory usage for small mixed L cases (where there are large off-block-diagonals). I've added tests to compare the dense vs. sparse values.


mariogeiger:
This is super cool.
Is the sparse version also faster?

To make it more sparse you can simply do
```
i = meshgrid
m = flatten values

i = i[m != 0]
m = m[m!=0]
```
it would be great to improve further to set to 0 the 1e-17 values in wigner3j

mariogeiger:
image
This is the distribution of log10 of the values in the `o3.wigner_3j(l1, l2, l3)` up to l=9.
Based on this I will set to 0 the entries smaller than 1e-14

mariogeiger:
Will be done in #89

PR #83, opened 2020-05-07
E3Conv as a message passing graph conv in pytorch geometric

blondegeek:
Here's my first attempt at an implementation of our convolutions as a message passing network in pytorch geometric. I'm not sure if timing wise it will be fast enough, but it seems to be a lot easier than dealing with rectangular neighbor lists.


e3nn/message_passing.py
```diff
@@ -0,0 +1,22 @@
+import torch
+from torch_geometric.nn import MessagePassing
+
+
+class E3Conv(MessagePassing):
+    def __init__(self, Kernel, Rs_in, Rs_out):
+        super(E3Conv, self).__init__(aggr='add')
```
mariogeiger, line 7:
```suggestion
        super().__init__(aggr='add')
```
for uniformity with the rest of the code

e3nn/message_passing.py
```diff
@@ -0,0 +1,22 @@
+import torch
+from torch_geometric.nn import MessagePassing
```
mariogeiger, line 2: add `torch-geometric` as dependency? @bkmi

bkmi: I suppose we could have it as an extra?

It seems like a lot to add for this one single use.

mariogeiger:
We will soon use pytorch-sparse as well

bkmi:
Are you talking about
https://github.com/rusty1s/pytorch_sparse
or 
https://pytorch.org/docs/stable/sparse.html?

mariogeiger:
rusty1s/pytorch_sparse

I think the other one is not working

PR #85, opened 2020-05-07
Post re-fork: Split SphericalTensor into SphericalTensor and FourierTensor

blondegeek:


blondegeek:
Same as https://github.com/e3nn/e3nn/pull/84 but after re-fork

PR #86, opened 2020-05-07
refork message passing E3Conv

blondegeek:


blondegeek:
Same as https://github.com/e3nn/e3nn/pull/83

mariogeiger:
Can you just add `torch_geometric` in the dependencies (i.e. in `pack.py`)?

PR #87, opened 2020-05-08
data helpers for neighbors and periodic neighbors

blondegeek:
This includes simple functions for creating neighbor lists and periodic neighbor lists using `ase` and `pymatgen` respectively. Also includes classes for storing this data in the pytorch geometric style inheriting from `torch_geometric.data.Data`.


mariogeiger:
I think we should put the file `e3nn/data_helpers.py` into `e3nn/point` and rename it with a more explicit name

tests/data_helpers_test.py
```diff
@@ -0,0 +1,24 @@
+# pylint: disable=not-callable, no-member, invalid-name, line-too-long, wildcard-import, unused-wildcard-import, missing-docstring
+import unittest
+
+import torch
+import e3nn.data_helpers as dh
+from e3nn import rs
+
+
+class Tests(unittest.TestCase):
```
mariogeiger, line 9: I think we should not use this `unittest` anymore and do like https://github.com/rusty1s/pytorch_sparse/blob/master/test/test_transpose.py

blondegeek: Do you want this addressed here or do you want us to rewrite all tests in this style in another PR?

blondegeek:
> I think we should put the file `e3nn/data_helpers.py` into `e3nn/point` and rename it with a more explicit name

I'm not sure what name would be more helpful but I'm fine with moving the file to `point`.

tests/data_helpers_test.py
mariogeiger: Just change it here and one day I will change them all in a separate PR

mariogeiger:
Move the `tests/data_helpers_test.py` as well :)
If you don't have a idea for the name we will go with that and maybe change later when we have an idea.

tests/data_helpers_test.py
blondegeek: Does that require changing something about how CI is run? I just use python3 to run tests with unittest but for pytest you have to run pytest...

PR #88, opened 2020-05-08
Tetris with torch geometric Data and Batch and data_helpers.

blondegeek:


mariogeiger:
I wait that we merge #87 before merging this one

mariogeiger:
Do you have an idea why I get the following error:
`TypeError: forward() got an unexpected keyword argument 'size'` ?

blondegeek:
Hmm... I did change `networks.py` to have `def forward(self, input, *args, **kwargs)` for `GatedConvNetwork`.

blondegeek:
This happens when you run `python3 examples/point/tetris_torch_geo.py`? The code works for me.

mariogeiger:
It works :-)

PR #90, opened 2020-05-12
Move SphericalTensor in folder `tensor`

mariogeiger:
- create folder `e3nn/tensor`

**SphericalTensor**
- remove `self.mul`
- remove `self.RadialModel`
- fix parity when applying operations like mul matmul etc
- remove division when computing the norm (has to be done by the user if he wants to)
- `self.signal` is shape `[..., channel]` and all method allows it

**FourierTensor**
- remove inheritance of `SphericalTensor` because it is not a `SphericalTensor`


e3nn/tensor/fourrier_tensor.py
```diff
@@ -0,0 +1,141 @@
+# pylint: disable=not-callable, no-member, invalid-name, line-too-long, missing-docstring, arguments-differ
```
blondegeek, line 1: Change  filename `fourrier_tensor.py` to `fourier_tensor.py`

PR #98, opened 2020-05-27
Various fixes to network and data_helpers

blondegeek:
Make `GatedParityConvNetwork` take general arguements (so it can be used with `e3nn.point.message_passing.Convolution`.

Correct `e3nn.point.data_helpers.neighbor_list_and_relative_vec` to give neighbors within distance D (previously was 2 * D + skin, set skin == 0.0 and divide sphere radius by 2).

### DIFF
--- a/e3nn/networks.py
+++ b/e3nn/networks.py
@@ -124,15 +124,9 @@ class GatedConvParityNetwork(torch.nn.Module):
         if 'n_norm' not in kwargs:
             kwargs['n_norm'] = N
 
-        if self.feature_product:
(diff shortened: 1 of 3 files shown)


blondegeek:
@mariogeiger Can we merge `GatedConvNetwork` and `GatedParityNetwork` with whether or not to use parity be implied by the type of `Rs_in` and `Rs_out` used?

mariogeiger:
The way `GatedBlockParity` and `GatedBlock` are constructed is very different. If we can merge something it's cool but in the same time these are the example networks that the user will read to start with, so if we can keep them simple it's good too.

PR #108, opened 2020-07-14
SmallConvolution -- sandwiches Kernel with two linear layers and resh…

blondegeek:
…apes input before kernel to make tensor product only between distinct groups

### DIFF
--- a/e3nn/point/message_passing.py
+++ b/e3nn/point/message_passing.py
@@ -43,22 +43,22 @@ class SmallConvolution(tg.nn.MessagePassing):
                              f"{lin_mul} % {tp_mul} = {lin_mul % tp_mul}")
         self.lin_mul = lin_mul
         self.tp_mul = tp_mul
-        
+
         self.Rs_in = rs.convention(Rs_in)
         self.Rs_out = rs.convention(Rs_out)
-        
+
         lps = [(l, p) for m, l, p in self.Rs_in]
         self.Rs_lin1 = [(1, l, p) for i in range(lin_mul) for l, p in lps]
         self.Rs_tp1 = [(1, l, p) for i in range(tp_mul) for l, p in lps]
-        
+
         lps = [(l, p) for m, l, p in self.Rs_out]
         self.Rs_lin2 = [(1, l, p) for i in range(lin_mul) for l, p in lps]
         self.Rs_tp2 = [(1, l, p) for i in range(tp_mul) for l, p in lps]
-        
+
         self.lin1 = KernelLinear(Rs_in, self.Rs_lin1)
         self.kernel = Kernel(self.Rs_tp1, self.Rs_tp2)
         self.lin2 = KernelLinear(self.Rs_lin2, Rs_out)
-        
+
     def forward(self, features, edge_index, edge_r, size=None, n_norm=1):
         """
         :param features: Tensor of shape [n_target, dim(Rs_in)]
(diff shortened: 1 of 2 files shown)


mariogeiger:
This is a sketch of how I understand what you did. In the middle I left many arrows for readability.
image

It looks like there is multiple paths that does the same job.
Maybe I would not play the multiplicity trick at the output.

blondegeek:
I think your diagram is correct, but just to make sure -- 1) There's an initial linear that makes the Rs have the same number of multiplicity across all orders and parities contained in the Rs_in. 2) Then before the convolution there's a reshape into different "groups" -- basically each group will interact with the same kernel (significantly reducing the number of paths in the Kernel which still allowing for a larger number of parameters to hold data) 3) Finally, the groups are flattened and then there's a final linear layer.

blondegeek:
Also turns out that the repo flake and my local flake differ on whether a linebreak should happen before or after a binary operator ([seems related to this](https://stackoverflow.com/questions/57074300/what-is-the-recommended-way-to-break-long-if-statement-w504-line-break-after-b)). Any idea why?

blondegeek:
One detail that I want to highlight is that `SmallConvolution` cannot be used in our general networks in `networks.py`. `Kernel` cannot be passed with `Rs_in` and `Rs_out` already provided as it need to be determined in the `__init__`. @mariogeiger do you have any preferences for structuring this type of function?

mariogeiger:
I would modify `GatedConvNetwork` such that it takes in argument a function `K, Rs_in x Rs_out -> Module`.

Then you can give him `lambda K, Rs_in, Rs_out: Convolution(K(Rs_in, Rs_out))` or `partial(SmallConvolution, lin_mul=2, tp_mul=2)`

blondegeek:
Anything you'd like me to add before I merge? I will have a follow up pull request for using `SmallConvolution` with / modifying `networks.py`.

mariogeiger:
What you do makes me think of Inverted Residuals.
Does the `lin1` and `lin2` operation really needs to be put inside the module?
I would put the linear operation outside and add non-lineraties between each layer.

mariogeiger:
If possible I would have one argument called `groups` like in [pytorch Conv2d](https://pytorch.org/docs/master/generated/torch.nn.Conv2d.html)

blondegeek:
> What you do makes me think of [Inverted Residuals](https://arxiv.org/abs/1801.04381).
> Does the `lin1` and `lin2` operation really needs to be put inside the module?
> I would put the linear operation outside and add non-lineraties between each layer.

I think for ease of use, I would but the linear operations inside the module -- whatever that module ends up being called if not `SmallConvolution`. I'm happy to consider adding non-linearities between the linear and kernel steps.

blondegeek:
> If possible I would have one argument called `groups` like in [pytorch Conv2d](https://pytorch.org/docs/master/generated/torch.nn.Conv2d.html)

Ooo. I didn't know `Conv2D` had a `groups` option. That's essentially what's happening with the middle layer (with the kernel). The reason why I didn't just make a Convolution that uses grouping is because I was concerned about how that would work with a nonlinearity like `GatedBlock`. The easiest way to stack `groups` with a given `Rs` is to have `Rs_all = Rs * groups` and I do not believe that this plays well with how `GatedBlock` returns the modified `Rs_out`.

blondegeek:
> Inverted Residuals

I had not heard about this before and I will have to read up before I'm able to say more.

mariogeiger:
I'm trying to implement `groups` option in `Kernel` https://github.com/e3nn/e3nn/pull/110

mariogeiger:
This image illustrate the idea of Inverted Residuals:
image

it's made of 1x1 convolutions (our Linear) and depthwise convolution (our Kernel with groups=mul_in=mul_out)

blondegeek:
See https://github.com/e3nn/e3nn/pull/117

PR #109, opened 2020-07-15
refactored CartesianTensor methods to give access to transformation matrix

blondegeek:


e3nn/tensor/cartesian_tensor.py
```diff
@@ -12,13 +12,19 @@ def __init__(self, tensor, formula=None):
...
+    def to_irrep_transformation(self):
         basis_change = o3.xyz_vector_basis_to_spherical_basis()
-        tensor = self.tensor
-        for i in range(self.tensor.dim()):
-            tensor = torch.tensordot(basis_change, tensor, dims=([1], [i])).transpose(0, i)
+        dim = self.tensor.dim()
+        change = torch.eye(3 ** dim).reshape(3 ** dim, *[3] * dim)
+        for j in range(3 ** dim):
+            for i in range(dim):
+                change[j] = torch.tensordot(basis_change, change[j], dims=([1], [i])).transpose(0, i)
```
mariogeiger, line 21:
```suggestion
        change = o3.direct_sum(*[o3.xyz_vector_basis_to_spherical_basis()] * dim)
```

blondegeek: Your change doesn't create the same matrix (and the tests fail). Basically it's a recursive block matrix multiplication. It has off diagonals. Perhaps I'm missing something?

mariogeiger: oh right. its not the correct matrix, try with `o3.kron` instead

blondegeek: Ah! So that's what `o3.kron` does! 🤣 Thanks!

PR #111, opened 2020-07-16
(draft) custom cuda implementation for message function 

L-sky:


e3nn/point/message_passing.py
```diff
@@ -29,3 +31,209 @@ def message(self, x_j, k):
         if k.shape[0] == 0:  # https://github.com/pytorch/pytorch/issues/37628
             return torch.zeros(0, k.shape[1])
         return torch.einsum('eij,ej->ei', k, x_j)
+
+
+class MinimalNetwork(torch.nn.Module):
```
blondegeek, line 36: This seems more fitting as a test rather than module in the repository as most likely folks will not want to use a MinimalNetwork.

e3nn/point/message_passing.py
```diff
@@ -29,3 +31,209 @@ def message(self, x_j, k):
...
+
+        Rs_in = representations[0]
+        Rs_out = representations[-1]
+
+        radial_model = partial(GaussianRadialModel, max_radius=3.2, min_radius=0.7, number_of_basis=10, h=100, L=3, act=swish)
+
+        self.real_spherical_harmonics = rsh
+        self.layer = TensorPassingLayer(Rs_in, Rs_out, radial_model, self.coupling_coefficients, self.coupling_coefficients_base_offsets, self.max_l_in, self.device)
+
+    def _find_max_l_out_in(self):
```
blondegeek, line 62: I'm a bit confused by this function. Why do you need to impose a max order on the user input Rs? I can understand why for the filters but `o3.selection_rule` has an argument `lmax` for setting the max filter spherical harmonics which handles this.

e3nn/point/message_passing.py
```diff
@@ -29,3 +31,209 @@ def message(self, x_j, k):
...
+                    tmp_base_pos += tmp_flat.shape[0]
+
+        self.coupling_coefficients_base_offsets = torch.tensor(tmp_base_offsets_list, dtype=torch.int32, device=self.device)
+
+    def forward(self, graph):
+        Y = self.real_spherical_harmonics(list(range(self.max_l + 1)), graph.rel_vec).contiguous()
+        return self.layer(graph.edge_index, graph.x, graph.edge_attr, Y)
+
+
+class TensorPassingLayer(torch_geometric.nn.MessagePassing):
```
blondegeek, line 108: Please add equivariance tests for this layer -- see `e3nn/tests/point/message_passing_tests.py` for examples.

e3nn/util/dataset/molecules.py
```diff
@@ -149,3 +149,135 @@ def download(self):
...
+
+
+def to_one_hot(ind_arr, n_classes):
+    one_hot = torch.zeros((len(ind_arr), n_classes))
+    one_hot[torch.arange(len(ind_arr)), ind_arr] = 1.
+    return one_hot
+
+
+# function that constructs graph given 
+def molecule_from_file_to_graph_sphere_connectivity_5_no_self_interactions(molecule_file, y):
```
blondegeek, line 173: Again, I think this is too specific a function to be included in the repo.

e3nn/util/dataset/molecules.py
```diff
@@ -149,3 +149,135 @@ def download(self):
...
+    return molecule_from_file_to_graph_base(molecule_file, y, connectivity_func=connections_within_sphere_5_no_self_interactions)
+
+def connections_within_sphere_5_no_self_interactions(mol):
```
blondegeek, line 176: Same here -- too specific.

e3nn/util/dataset/molecules.py
```diff
@@ -149,3 +149,135 @@ def download(self):
...
+    return connections_within_sphere(mol, r_cut=5.0, self_interaction=False)
+
+def connections_within_sphere(mol, r_cut, self_interaction):
```
blondegeek, line 179: Why not use the functions from `e3nn.point.data_helpers.DataNeighbors`? They do exactly this.

e3nn/util/dataset/molecules.py
```diff
@@ -149,3 +149,135 @@ def download(self):
 
         self.files = sorted(glob.glob(os.path.join(xyz_path, "*.xyz")))
         assert len(self.files) == 133885
+
```
blondegeek, line 152: I think these changes should be removed from this pull request since they don't have to do with the cuda implementation of the message_passing convolution.

blondegeek:
Hi @L-sky, 

Thanks for making the pull request. It will be exciting to have a CUDA kernel for the message_passing version of the e3nn convolutions.

I took a look and I think the first thing I would request that when you are ready to make the final pull request (since I know this is still being drafted) please break the pull request into smaller independent pieces that we can layer to make all the changes. That way I can better ensure that I can do a proper code review.

In general, I see several functions that I don't think need to be in the repo because they are too specific, just keep an eye out for which functions are general enough to be used by other use cases  versus for a specific use case.

I think the `molecule.py` file should be moved to a different pull request and also please look at the `DataNeighbors` module and make sure that we are not bloating the e3nn repo with redundant code. If `DataNeighbors` is missing some functionality, then it should be modified, but from what I could tell I think the functionality is very similar if not identical.

Let me know if you'd like me to take a look at anything. Thanks!

e3nn/util/dataset/molecules.py
L-sky: there are two reasons: 

1) DataNeighbor additionally uses `aes` package, while I keep it to `pymatgen` only - I prefer to minimize number of additional packages.

2) DataNeighbors creates Data object with `edge_attr` for all molecules at the same time. In order to keep memory consumption at bay, I move creation of `edge_attr` to the runtime (transform function), which is only possible for Dataset class, not list of Data objects.

e3nn/point/message_passing.py
L-sky: True. This is a temporary class for development purposes that I used to debug calls to cuda extensions, whether gradients are computed and such.

e3nn/point/message_passing.py
L-sky: It does not impose anything, it literally finds maximal values of `l_out` and `l_in` used across the network. This function is needed to determine which coupling coefficients to store, as well as to construct offsets for cuda calls.

mariogeiger:
Hi Kostiantyn, this is a lot of code to review !
Can you summarize what you did?

L-sky:
@mariogeiger Just to be on the same page, this PR is not to be merged. It is here, so that it is easier to communicate where development is going on my part. Once it stabilizes (settled how to handle installation), I'll create separate PRs with independent chunks. 

As is in the current state, it is mostly:
1) Tensor passing layer, network  - cuda kernels, python wrappers and infrastructure around it to fit into message passing framework. 
2) Real spherical harmonics - cuda kernel for normalization + fix python wrapper.

mariogeiger:
I close PR on master

PR #114, opened 2020-07-17
depthwise convolution

mariogeiger:


blondegeek:
See https://github.com/e3nn/e3nn/pull/117

PR #121, opened 2020-07-30
DepthwiseConvolutionParity

blondegeek:


blondegeek:
`GatedBlockParity` requires reordering `Rs_in` to be broken up into scalars and nonscalars which has undesirable side effects for grouping. I think my solution could be replaced with a rewriting of `GatedBlockParity` such that it behaves more like `GatedBlock`.

mariogeiger:
We should do a video call such that you can explain me what is going on. Shall I merge it already now and we discuss it later?

blondegeek:
I think merging would be good just to maintain for now -- I think this
should be simplified with some future verison of LearnableTensorProduct /
WeightedTensorProduct. I suspect we'll have a better sense of how to
modularize as we think more about two atom output tensors.

PR #122, opened 2020-07-31
DepthwiseNetwork

blondegeek:
This is a network that uses DepthwiseConvolutionParity.
There's currently something wrong with it because it will often output all zeros for random input upon initialization. I do use `allow_unused_inputs = True` for the Kernel which might be causing a shrinking of the variance, but I'm not sure. When the network does give a non-zero output, it is equivariant.


blondegeek:
https://github.com/e3nn/e3nn/pull/121

PR #129, opened 2020-08-25
Variable networks

mister-bailey:
GatedConvNetwork already allows you to specify layer-by-layer representation details. However, GatedConvParityNetwork only lets you specify lmax and multiplicity for the hidden layers. We've been using the class VariableParityNetwork in our project to build GatedConvParityNetworks with variable lmaxes and multiplicities. I've put it in a separate file, variable_networks.py, but something like this could very well be brought into networks.py if people are interested.

ALSO I have put batch norms in between layers in VariableParityNetwork, as an initialization option. These weren't in the existing network classes, but could easily be added there as well.


mariogeiger:
Can you just fix the linter such that I can accept this PR?

JoshRackers:
What ever happened with this? Was there a problem besides the linter?

mariogeiger:
At the time it was just the linter the problem. But since I am creating a more general architecture `MLNetwork` that this kind of architecture can specialize. New PR are welcome and I am happy to help with code writing

PR #130, opened 2020-08-25
Fix for windows

mister-bailey:
This is a necessary fix for e3nn to work on windows. There is only one file where there's a problem: util/cache_file.py

Windows doesn't have fcntl, or any analog that I'm aware of. Therefore I test os.name before using it. On windows, we don't really get an inter-process filesystem lock. Beware.

### DIFF
--- a/e3nn/util/cache_file.py
+++ b/e3nn/util/cache_file.py
@@ -1,11 +1,11 @@
 '''
 Cache in files
 '''
+import glob
+import os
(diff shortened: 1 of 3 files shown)


mister-bailey:
Yikes. This branch has picked up all my changes on the variable_networks branch! I will change this...

mariogeiger:
How is it going with this PR?

PR #135, opened 2020-09-20
add doc strings to some network classes and remove feature_product

blondegeek:
@mariogeiger let me know how you want to format long lines for doc strings.

### DIFF
--- a/e3nn/networks.py
+++ b/e3nn/networks.py
@@ -21,16 +21,16 @@ class GatedConvNetwork(torch.nn.Module):
     This network architecture should be used if you only want to be equivariant to rotations and translations but not inversion. GatedBlock acts on irreps of SO(3), meaning it does not include parity. If you want mirror symmetry and inversion, use GatedConvParityNetwork which handles the full irreps of O(3).
 
     Args:
-        Rs_in ([rs.TY_RS_STRICT]): Representation list of input data. Parity must be omitted or set to 0.
-        Rs_hidden ([rs.TY_RS_STRICT]): Representation list of intermediate data. Parity must be omitted or set to 0.
-        Rs_out ([rs.TY_RS_STRICT]): Representation list of output data. Parity must be omitted or set to 0.
-        lmax ([int > 0]): Maximum L used for spherical harmonic in kernel.
+        Rs_in (rs.TY_RS_STRICT): Representation list of input data. Parity must be omitted or set to 0.
+        Rs_hidden (rs.TY_RS_STRICT): Representation list of intermediate data. Parity must be omitted or set to 0.
+        Rs_out (rs.TY_RS_STRICT): Representation list of output data. Parity must be omitted or set to 0.
+        lmax (int > 0): Maximum L used for spherical harmonic in kernel.
         layers (int, optional): Number of convolution + GatedBlock layers. Defaults to 3.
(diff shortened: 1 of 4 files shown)


mariogeiger:
as you wish. We don't need `feature_product` anymore?

blondegeek:
It’s too computationally expensive to be practical — ideally it would be
replaced by LearnableTensorProduct or DepthwiseConvolution or
WeightedTensorProduct — which we should probably figure out how to reduce
to a simpler set of classes.

e3nn/networks.py
```diff
@@ -16,9 +16,49 @@
...
 class GatedConvNetwork(torch.nn.Module):
+    """A basic network architecture that alternates between Convolutions and GatedBlock nonlinearities.
+
+    This network architecture should be used if you only want to be equivariant to rotations and translations but not inversion. GatedBlock acts on irreps of SO(3), meaning it does not include parity. If you want mirror symmetry and inversion, use GatedConvParityNetwork which handles the full irreps of O(3).
+
+    Args:
+        Rs_in ([rs.TY_RS_STRICT]): Representation list of input data. Parity must be omitted or set to 0.
+        Rs_hidden ([rs.TY_RS_STRICT]): Representation list of intermediate data. Parity must be omitted or set to 0.
+        Rs_out ([rs.TY_RS_STRICT]): Representation list of output data. Parity must be omitted or set to 0.
+        lmax ([int > 0]): Maximum L used for spherical harmonic in kernel.
```
mariogeiger, line 27: why the notation `([type])`? It confuse me because I read it `a list of objects of type type`

e3nn/networks.py
```diff
@@ -176,7 +272,8 @@ def forward(self, input, *args, **kwargs):
             output = conv(output, *args, **kwargs)
             shape = list(output.shape)
             # Split multiplicities into new batch
-            output = output.reshape(shape[:-1] + [self.mul, (self.lmax + 1) ** 2])
+            output = output.reshape(
+                shape[:-1] + [self.mul, (self.lmax + 1) ** 2])
```
mariogeiger, line 276: was this line really too long??

mariogeiger (commented):
This is super nice that you added all those docstrings ! It's now super good entry point for new users :+1:

e3nn/networks.py
blondegeek: Ah, it should just be (type) -- I was using autogenerated doc strings from VS Code and I forgot to delete the extra parenthesis. I will fix and push.

blondegeek:
I'm very confused about the error pytest is returning -- its for code that isn't in this commit
```
    def test_rsh_backwardable():
        lmax = 10
        Rs = [(1, l) for l in range(lmax + 1)]
    
        xyz = torch.tensor([
            [0., 0., 1.],
            [1.0, 0, 0],
            [0.0, 1.0, 0],
        ], requires_grad=True, dtype=torch.float64)
>       assert torch.autograd.gradcheck(partial(rsh.spherical_harmonics_xyz, Rs), (xyz,), check_undefined_grad=False)
E       TypeError: gradcheck() got an unexpected keyword argument 'check_undefined_grad'
```

mariogeiger:
haha it's a feature only available in pytorch 1.6.0 but somehow the CI script is not up to date in this branch

mariogeiger:
We can marge

PR #139, opened 2020-10-14
relax Rs loose type

mariogeiger:
:boom: Don't merge if you are not sure ! It is only a proposal, I don't know if it's a good idea

Allow
```
Rs = (1, 1)
```
to declare a vector

Advantages:
- This works `rs.cut(features, *Rs)`

Cons:
- `[1, 2]` is different than `(1, 2)`

### DIFF
--- a/e3nn/rs.py
+++ b/e3nn/rs.py
@@ -17,7 +17,7 @@ from e3nn import o3, perm
 from e3nn.util.default_dtype import torch_default_dtype
 from e3nn.util.sparse import get_sparse_buffer, register_sparse_buffer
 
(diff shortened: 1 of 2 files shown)


blondegeek:
I actually have no idea what `rs.cut` does from looking at the code. I know this is not part of the original pull request, but can you add a docstring with an example?

blondegeek:
Is there a specific use case for this relaxed typing where it's been inconvenient to use the full Rs? Or is this suppose to give a complementary option to the case of just Ls?

mariogeiger:
I had the idea of doing that while reading your notebook
```
for (mul, l, p), x in zip(rs.convention(Rs), rs.cut(order_parameters, *Rs)):
    print("({}, {}, {}) => {}".format(mul, l, px))
```

PR #144, opened 2020-11-13
Update tetris.py

sophiaas:
Fixes the labels on the tetris datapoints. I noticed the errata after visualizing the dataset.


mariogeiger:
Thanks 😁

PR #151, opened 2020-12-01
Clarify Rs_sh / sh in tetris example.

bdice:
Hi, I have a question and a potential solution for my question in this PR.

https://github.com/e3nn/e3nn/blob/4c12f848c07e5d1b149ff9a68f525c7e2e6f885b/examples/point/tetris_torch_geo.py#L49
https://github.com/e3nn/e3nn/blob/4c12f848c07e5d1b149ff9a68f525c7e2e6f885b/examples/point/tetris_torch_geo.py#L65

I was modeling my code on this `tetris_torch_geo.py` example.
I tried to change `Rs_sh` in my code, but the tensors were not correctly-sized until I made `sh` match.
I am guessing that `sh` is defined here (and not left as `None`, which is handled in `WTPConv.forward` based on `Rs_sh`) because it's being pre-calculated and re-used in each convolution layer.

If my understanding is correct, I'd like to add a code comment that mentions the use of pre-computation, and perhaps use `Rs_sh` directly in the expression for `sh` so that the choice of representations can be
changed in just one place (`Rs_sh`) and the correspondence is clearer.

Let me know what you think. Thanks!

### DIFF
--- a/examples/point/tetris_torch_geo.py
+++ b/examples/point/tetris_torch_geo.py
@@ -41,7 +41,7 @@ def get_dataset():
     return tetris, labels
 
 
-def forward(f, shapes, Rs_sh, lmax, device):
+def forward(f, shapes, Rs_sh, device):
     r_max = 1.1
     x = torch.ones(4, 1)
     batch = Batch.from_data_list([DataNeighbors(x, shape, r_max, self_interaction=False) for shape in shapes])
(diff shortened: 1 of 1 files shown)


mariogeiger:
Good!
But then no need to have `lmax` as argument ?

bdice:
@mariogeiger Good catch. Fixed.

PR #152, opened 2020-12-01
Rewrite docstrings and add missing parameters for neighbor list methods.

bdice:
I improved some docstrings in `data_helpers.py` to clarify some information that wasn't clear to me about neighbor finding.

I'm guessing the standard is Google-style docstrings (`Args:`) based on the rest of the code base, but there are some Sphinx-style `:param name:` mixed in.
Let me know if you prefer another way.

Thanks!


mariogeiger:
No preference, do you have a suggestion?

PR #153, opened 2020-12-01
Pin pip to 20.2.4.

bdice:
I noticed my tests failed on #151 and #152. This PR (hopefully) fixes the CI issues caused by pip 20.3's new, stricter resolution logic.

See this upstream issue where I commented for more details:
https://github.com/rusty1s/pytorch_geometric/issues/1876


bdice:
Unfortunately the only solution I see for now is to install dependencies with pip 20.2.4. Pinning the versions won't work yet -- see my edited comment on the issue linked above.

mariogeiger:
Thanks!

PR #155, opened 2020-12-01
Ignore shared libraries built in place.

bdice:
I built e3nn with CUDA support "in-place", meaning that it puts the shared library `cuda_rsh.cpython-38-x86_64-linux-gnu.so` in the `e3nn` directory. This PR makes git ignore the built shared libraries in the source directory.


mariogeiger:
:+1:

PR #162, opened 2020-12-21
Unified `DataNeighbors` using newer ASE neighbor lists

Linux-cpp-lisp:
This PR updates the `DataNeighbors` data helpers to improve a few things:

- Use the newer linear-scaling [ASE neighbor lists](https://wiki.fysik.dtu.dk/ase/ase/neighborlist.html) instead of the older, quadratic ones (which are used by default in the `ase.neighborlist.NeighborList` class)
- Use ASE consistently for both periodic and non-periodic systems (rather than using `pymatgen` for periodic systems)
- Expose ASE's support for partial periodic boundary conditions such as surfaces that are periodic in two-of-three cell vectors
- Avoid the allocation of intermediate `ase.Atoms`/`pymatgen.Structure` objects to improve dataloading efficiency
- Add `DataNeighbors.from_ase`
- Adds and updates relevant tests

I've concentrated all the neighbor finding into `DataNeighbors` and `_neighbor_list_and_relative_vec` by adding `cell` and `pbc` parameters (following ASE conventions); I've also rewritten `DataPeriodicNeighbors` and `_neighbor_list_and_relative_vec_lattice` as compatibility wrappers for the former. Whether those should be marked as deprecated is up to you. Also the cell is added to a `DataNeighbors` object as a property named `lattice` even though it's passed in as `cell` for backwards compatibility: I could change this to be consistent but I wanted to start out not breaking things.

The only break in compatibility should be the order of edges in periodic systems, since ASE and pymatgen return them in different orders, but that shouldn't matter. 

Thanks!

### DIFF
--- a/e3nn/point/data_helpers.py
+++ b/e3nn/point/data_helpers.py
@@ -42,6 +42,7 @@ class DataNeighbors(tg.data.Data):
             **kwargs
         )
 
+
 class DataPeriodicNeighbors(DataNeighbors):
     def __init__(self, x, pos, lattice, r_max, self_interaction=True, **kwargs):
         super().__init__(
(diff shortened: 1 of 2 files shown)


mariogeiger:
Thanks a lot for the PR!
It looks good

blondegeek:
This is great! Thanks Alby!

PR #185, opened 2021-01-27
Norm-based nonlinearity

Linux-cpp-lisp:
Add a norm-based nonlinearity to new e3nn.

See [`NormActivation`](https://github.com/e3nn/e3nn/blob/master/e3nn/non_linearities/norm_activation.py) in old e3nn.

Still TODO:
- [X] Equivariance unit test 
- [X] Docs
- [X] Check correctness of given scalar nonlinearity (are scalar nonlinearities that take on negative values valid for equivariance?)


e3nn/nn/normact.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self,
+                 irreps_in,
+                 scalar_nonlinearity,
+                 normalize=True,
+                 bias=False,
+                 epsilon=1e-8,
+                 **kwargs):
+        super().__init__()
+        self.irreps_in = irreps_in
+        self.irreps_out = irreps_in
```
mariogeiger, line 43: I'm use to call `o3.Irreps(irreps_in)` in order to allow the user to do `NormActivation("0e + 1e")`

e3nn/nn/normact.py
```diff
@@ -0,0 +1,87 @@
...
+        self.norm = Norm(irreps_in, **kwargs)
+        self.scalar_nonlinearity = scalar_nonlinearity
+        self.epsilon = torch.as_tensor(epsilon)
+        self.normalize = True
```
mariogeiger, line 47: `self.normalize = normalize`?

e3nn/nn/normact.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Returns
+        -------
+        `torch.Tensor`
+            tensor of shape ``(..., irreps_out.dim)``
+        '''
+        norms = self.norm(features)
+
+        # Apply nonlinearity
+        if self.normalize:
```
mariogeiger, line 74: What about `if self.epsilon is not None`? Just an idea like that... I let you juge

e3nn/nn/normact.py
```diff
@@ -0,0 +1,87 @@
...
+            epsilon_norms = norms.clone()
+            epsilon_norms[epsilon_norms < self.epsilon] = self.epsilon
+            nonlin_arg = epsilon_norms
+            if self.bias:
+                nonlin_arg = nonlin_arg + self.biases[None, :]
```
mariogeiger, line 79: Why `[None, :]`?

mariogeiger (commented):
Comments on the fly

e3nn/nn/normact.py
Linux-cpp-lisp: Since it's an explicit change to the functional form of the nonlinearity, it seemed to me like it should have an explicit parameter.

e3nn/nn/normact.py
Linux-cpp-lisp: Good catch thanks

e3nn/nn/normact.py
Linux-cpp-lisp: removed

Linux-cpp-lisp:
This should be ready to go now.

Note that I merged https://github.com/e3nn/e3nn/pull/188 into this to be able to test equivariance, so that should probably be merged before this is.

I'm also still not completely sure about the 3rd point in the TODO, even though the numerical equivariance test with `tanh` and parity inversions and biases does pass. The _direction_ of the vectors does flip when `tanh` is used with biases, but I changed the test to allow anything that stays on the same line as the input regardless of direction.

Issue #184, opened 2021-01-26
[Enhancement] About the usage of __new__ and superclasses


jnooree:
Hello, I'm trying to use this library for my small machine learning project. While reading the source codes figuring out how to use this, I have noticed some enhancements could be made.

First, I've found out some `__new__()` methods are being used in an unusual way.
Python `__new__` magic method takes a type (or a class) for its first argument, and it is normally represented as `cls`. I didn't analyzed all of the sources, but some of them are using `self` as the first argument, which might cause confusion that it is an actual instance.

Second, I've seen that some superclass names are hard-coded into the `__new__` methods. I guess `super()` call would do the job with more flexibility, if the exact superclasses are not directly required.

Thanks.

mariogeiger:
Hi, thanks for the tips! It's the first time I used `__new__` 😅. 
Would you be motivated to write a PR? Otherwise I can also look at it.

jnooree:
If you don't mind some delay, I might be able to make the PR done right after this weekend (GMT +9:00).

PR #186, opened 2021-01-28
Update __new__ methods

jnooree:
- Change the first argument to cls, following the conventions
- Change explicit superclass access to super() call

This will close e3nn/e3nn#184.


mariogeiger:
Thanks!

PR #188, opened 2021-01-28
Refactored tests and equivariance tests

Linux-cpp-lisp:
This PR has a few main features:

1. The global test fixture `float_tolerance` that runs every test with a pytorch default dtype of both `float32` and `float64`, and that passes a corresponding error tolerance value through to any tests that need one.
2. The `equivariance_error` utility function that, given some callable and irreps describing its inputs and outputs, runs a configurable number of random equivariance trials and returns the largest componentwise error across them.
3. The `assert_equivariant` fixture, which gives any tests that want it a function that takes a callable and its input/ouput irreps and confirms that its `equivariance_error` is below `float_tolerance`.

Using this, I've replaced most of the obvious equivariance tests with calls to `assert_equivariance`, as well as adding a lot of equivariance tests that didn't exist before. Everything passes — it's all equivariant with both `float32` and `float64` :tada: 

A few notes:
- `equivariance_error` also runs tests for parity transforms, which the old tests did not do
- I mapped hardcoded tolerances to `float_tolerance` as follows: 
    - `1e-10` in tests previously with a `float64` default dtype => `float_tolerance`
    - `1e-9` in the same => `10*float_tolerance`
    - `1e-5` in the same => `sqrt(float_tolerance)`
    - Other more forgiving thresholds (ex. `1e-3`) were left hardcoded
- Some of the group theory tests have _terrible_ error with `float32`. Since these aren't actual model tests I marked those as `xfail`s rather than significantly relaxing the threshold. Those are:
    - `cartesian_spherical_harmonics.test_recurrence_relation`
    - `group_test.test_representation`
    - `group_test.test_is_group`

Open questions:
- How many random equivariance trials should we do for each callable? I set this at 2 right now in hopes of catching any random behavior.
- I set the `float_tolerance` to 1e-10 for `float64` and 1e-4 (!) for `float32`; this is necessary to get things to pass. I suspect that the dramatic difference is just due to compounding errors. It could probably be tightened at the cost of explicitly introducing another fixture, say `generous_float_tolerance`, that gives more wiggle room when running on `float32` to tests that know they involve a lot of compounding error. Thoughts?


e3nn/util/equivariance.py
```diff
@@ -0,0 +1,122 @@
...
+        irreps_in = [irreps_in]
+    elif isinstance(irreps_in, list) or isinstance(irreps_in, tuple):
+        irreps_in = [i if i == 'points' else o3.Irreps(i) for i in irreps_in]
+    else:
+        irreps_in = [o3.Irreps(irreps_in)]
+
+    if isinstance(irreps_out, o3.Irreps):
+        irreps_out = [irreps_out]
+    elif isinstance(irreps_out, list) or isinstance(irreps_out, tuple):
+        irreps_out = [i if i == 'points' else o3.Irreps(i) for i in irreps_out]
```
mariogeiger, line 48: Maybe cartesian is a better naming?
What do you think @blondegeek?

e3nn/util/equivariance.py
```diff
@@ -0,0 +1,122 @@
...
+
+        for parity_k in parity_ks:
+            args = [
+                torch.randn(point_shape) if irreps == 'points' else irreps.randn(*arg_shape)
+                for irreps in irreps_in
+            ]
+            # Build a rotation matrix for point data
+            rot_mat = o3.rand_matrix()
+            D_params = (rot_mat, parity_k)
+            # TODO: is parity needed for rot_mat?
```
mariogeiger, line 83: I think it is correct as you did

e3nn/util/equivariance.py
```diff
@@ -0,0 +1,122 @@
...
+            rot_mat *= (-1)**parity_k
+
+            # Evaluate the function on rotated arguments:
+            rot_args = [
+                (a @ rot_mat.T) if irreps == 'points' else (a @ irreps.D_from_matrix(*D_params).T)
```
mariogeiger, line 88: This makes me think that `D_from_matrix` should be able to get the parity from the determinant of the rotation matrix, no need to give the parity as a second argument...
I did that "mistake" because it is not the case for euler angles and quaternions which cannot (asaik) encode the parity

tests/o3/reduce_tensor_test.py
```diff
@@ -32,23 +31,24 @@ def test_reduce_tensor_antisymmetric_L2():
...
+
+    assert_equivariant(tp, irreps_in=tp.irreps_in, irreps_out=tp.irreps_out)
 
 
-def test_reduce_tensor_elasticity_tensor():
+def test_reduce_tensor_elasticity_tensor(assert_equivariant):
     tp = o3.ReducedTensorProducts('ijkl=jikl=klij', i='1e')
     irreps = tp.irreps_out
     assert irreps.dim == 21
+    assert_equivariant(tp, irreps_in=tp.irreps_in, irreps_out=tp.irreps_out)
```
mariogeiger, line 43: Beautiful!

tests/o3/tensor_product_sub_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_id(assert_equivariant):
+    irreps_in = o3.Irreps("1e + 2e + 3x3o")
+    irreps_out = o3.Irreps("1e + 2e + 3x3o")
+    m = Identity(irreps_in, irreps_out)
+    print(m)
+    m(torch.randn(irreps_in.dim))
+
+    assert_equivariant(m)
```
mariogeiger, line 29: Amazing! :-)

mariogeiger (approved):
Thanks a lot, it is so much cleaner like that!

I think testing equivariance only once is fine, we don't want the test to take ages.

mariogeiger:
To check the linter locally you can run 
```flake8 . --count --select=E9,F63,F7,F82,E251,F811,E302,E261,E231,F401,F841,W293,W291,W605,W292,W391,E128,E305,W503,W504,E303,E125,E121,E123,E203,F541 --show-source --statistics```
or configure your editor to run flake8

e3nn/util/equivariance.py
mariogeiger: https://github.com/e3nn/e3nn/issues/189

Linux-cpp-lisp:
I did finally figure out how to get the linter working in my editor... that's a useful command too, thanks! And sorry for all the linter commits :eyes:

Linux-cpp-lisp:
Merged as part of https://github.com/e3nn/e3nn/pull/185

PR #191, opened 2021-02-01
Make equivariance tests available to outside code

Linux-cpp-lisp:
- Makes `equivariance_error`, `assert_equivariance` available to client code for its tests
- Improve reporting of equivariance error to specify what type of transform caused a problem (was there parity, was there a translation)
- Added equivariance testing to the docs


mariogeiger:
This is super cool!

There is one point I want to change.
Let's distinguish points from relative positions.
1. Points do transform under translation
2. Relative positions don't

I propose `point` and `cartesian`
or `point` and `vector`
or `point` and `relative_position`

Linux-cpp-lisp:
That's a great point @mariogeiger: I think `point` and `vector` is probably clearest and will make that change.

PR #193, opened 2021-02-03
Initial TorchScript Support

Linux-cpp-lisp:
### TorchScript support
This is a preliminary PR for TorchScript support, which is necessary in many cases for deploying e3nn-based models in production. Because e3nn uses a lot of python for model architecture — subclassing, `Irreps`, etc. — that isn't supported by TorchScript, I'm focusing on tracing with some helper functions in TorchScript (`@torch.jit.script`) where necessary.

So far:
- [X] TorchScript `broadcast_tensors`
- [X] JITable `TensorProduct` and subclasses
- [X] Testing tool for JIT traceability
- [X] `FullyConnectedNet`

This involves some changes to `TensorProduct`, so it probably needs to be looked over carefully.

### TODO
Go through other critical modules and functions and test + implement TorchScript support. At a minimum before merging I think that means doing:
- [ ] `NormActivation`
- [ ] `Gate`
- [ ] Spherical harmonics expansion

More modules can probably be done piecemeal as needed.

### A problem
In general, making a module TorchScript compatible seems to involve either: wrapping some simple architecture-related asserts and input reshaping in TorchScript (`TensorProduct`) **or** scripting the whole module (ex. `BatchNorm`).

In the second case, this means that we need a way to script certain submodules while tracing their parents (instead of following the trace into the submodule). Otherwise, there's no way to trace actual production modules like `gate_points_2101.Network`. As far as I can tell this would require either:
- Always scripting them: we're already loading the compiler for `TensorProduct`, so this isn't necessarily a problem. It is, however, a pain because then you always have to say `self.bn = torch.jit.script(BatchNorm(...))` instead of `self.bn = BatchNorm(...)`.
- Added a thin e3nn wrapper over `torch.jit.trace` that first recursively checks submodules for, say, `@e3nn.jit.script_when_tracing` and scripts them first. There actually is a [`@torch.jit.script_if_tracing`](https://pytorch.org/docs/master/generated/torch.jit.script_if_tracing.html), but it doesn't work on modules.

Would appreciate if anyone has thoughts on this problem.


tests/o3/tensor_product_sub_test.py
```diff
@@ -28,6 +26,7 @@ def test_id():
     m(torch.randn(irreps_in.dim))
 
     assert_equivariant(m)
+    assert_jit_trace
```
mariogeiger, line 29: Evaluation missing?

mariogeiger (commented):
Nice!! Looking at the code seems very nice. I like the broadcast function into the jit code. 
We should chat before or during the meeting because I don't understand the problems you mentioned.

e3nn/o3/tensor_product.py
```diff
@@ -478,12 +514,15 @@ def main(x2: torch.Tensor, ws: List[torch.Tensor], w3j: List[torch.Tensor]) -> t
...
-        s = 4 * " "
-        code_out += f"{s}return out"
-        code_right += f"{s}return out"
+        code_out += f"{s}return out.reshape(outsize)"
+        code_right += """
+        return out.reshape(outsize)
+"""
 
         self.code_out = code_out
+        self.main_out = eval_code(self.code_out).main
```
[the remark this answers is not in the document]
Linux-cpp-lisp: Resolved in latest push & added test

tests/o3/tensor_product_sub_test.py
Linux-cpp-lisp: Fixed, good catch thanks

mariogeiger:
You even added a test for the deepcopy :heart:

PR #197, opened 2021-02-10
TorchScript JIT support for e3nn

Linux-cpp-lisp:
## Description
PyTorch provides two ways to compile code into TorchScript: tracing and scripting. Tracing follows the tensor operations on an example input, allowing complex Python control flow _if_ that control flow does not depend on the data itself. Scripting compiles a subset of Python directly into TorchScript, allowing data-dependent control flow but very limited Python features.

This is a problem for e3nn, where many modules — most of all `TensorProduct`s — use significant Python control flow based on `Irreps` as well as features like inheritance that are incompatible with scripting. Other modules, however, like `Gate` or `Convolution`, contain important but simple data-dependent control flow. Thus `Convolution` needs to be scripted, even though it contains a `TensorProduct` that has to be traced.

PyTorch allows scripting and tracing to be mixed for precisely this reason but provides no clean way of doing so automatically: users would be required to carefully trace and script various submodules in the `__init__` methods of their models. This would be extremely error prone and require significant and complex code duplication for the construction of tracing inputs.

To fix this problem, this PR implements a wrapper for `torch.jit` — `e3nn.util.jit` — that recursively and automatically compiles submodules according to directions they provide. Using the `@compile_mode` decorator, modules can indicate whether they should be scripted, traced, or left alone. Infrastructure is also put in to allow automatically generating tracing inputs for most modules, and those with more complex requirements can define `_make_tracing_inputs` to override the built-in behavior.

As a result, the `gate_points_2101` model can be compiled down to TorchScript with only two changes to its source code:
 - marking `Convolution` with `@compile_mode('script')`
 - allowing dictionary inputs in addition to `torch_geometric.data.Data` inputs (`torch.jit.trace` requires tracing inputs to be of simple types)

## Further work

I didn't do every element of the library, in particular `BatchNorm` and `S2Act` are not yet supported. There's no reason why it wouldn't be possible, but those two are both more specialized and more complex and should be looked at in separate PRs.

If this PR gets approved, I will also take a first stab at writing a page in the docs on how to compile e3nn models.

## Motivation and Context
TorchScript models allow deployment of e3nn models without Python and the e3nn library, as well as allowing Python-less model evaluation and PyTorch's compute graph optimization.

## How Has This Been Tested?
JIT tests have been added for most modules using `assert_auto_jitable`.
Existing tests pass.
A full JIT integration test for `gate_points_2101_test.py` that saves and then loads a TorchScript model.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds or improves functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:
- [X] I have read the [**CONTRIBUTING**](https://github.com/e3nn/e3nn/blob/main/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] I have updated the documentation (if relevant).
- [X] I have added tests that cover my changes (if relevant).
- [X] All new and existing tests passed.
- [X] I have updated the [Changelog](https://github.com/e3nn/e3nn/blob/main/ChangeLog.md).


tests/nn/models/gate_points_2101_test.py
```diff
@@ -27,13 +36,72 @@ def test_gate_points_2101(float_tolerance):
...
+# Needed for the TorchScript kernels for scatter and radius_graph
+import torch_scatter
+import torch_cluster
+f = torch.jit.load('{tmpdir}/model.pt')
+d = torch.load('{tmpdir}/dat.pt')
+out = f(d)
+torch.save(out, '{tmpdir}/out.pt')
+""")
+        # Run
+        proc_res = subprocess.run(['python', tmpdir + '/code.py'])
```
mariogeiger, line 103: Maybe there is a way to obtain the currently used python interpreter

mariogeiger (commented):
Impressive work!
I saw that you have some `TODO`.
Tell me when you think it is ready for merge?

tests/nn/models/gate_points_2101_test.py
Linux-cpp-lisp: Resolved in e22a35b

e3nn/nn/gate.py
```diff
@@ -100,21 +104,42 @@ def key(mul_ir):
...
+                            f"{s}out[{out_i}][..., {i_out}:{i_out + d}] = x[..., {i_in}:{i_in + d}]"
+                        )
                         i_in += d
                     i_out += d
-        return outs
+
+        code_out.append(f"{s}return out")
+
+        self.code_out = "\n".join(code_out)
+        self._compiled_main_out = eval_code(self.code_out).main
```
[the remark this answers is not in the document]
Linux-cpp-lisp: Resolved in eb4fe4e 

I added a small little mixin class, `CodeGenMixin`, that manages eval code and copying stuff for classes that have dynamic code generation. It has the trick for copying and is now a mixin for `TensorProduct` and `_Sortcut`.

Linux-cpp-lisp:
> Impressive work!
> I saw that you have some `TODO`.
> Tell me when you think it is ready for merge?

Thanks Mario! If you're good with the PR, I'll write up a first attempt at a docs page, upload the changelog, and this should then be ready to merge. I think `BatchNorm`, and `S2Act` can wait for later PRs.

mariogeiger:
I just got two problems, one warning and an error.

1. *warning*
```python
import torch
from e3nn import o3
from e3nn.util.jit import trace
from e3nn.nn.models.gate_points_2101 import Network

f = Network(
    irreps_in=None,
    irreps_hidden='100x0e+10x1e',
    irreps_out='0e+1e',
    irreps_node_attr=None,
    irreps_edge_attr='0e+1e',
    layers=4,
    max_radius=1.0,
    number_of_basis=4,
    radial_layers=1,
    radial_neurons=100,
    num_neighbors=4.0,
    num_nodes=4.0,
)
f(dict(pos=torch.randn(4, 3)))
f_traced = trace(
    f,
    example_inputs=(dict(pos=torch.randn(4, 3)),),
#     check_inputs=[(random_graph(),) for _ in range(3)]
)
```
output:
```
/home/mario/anaconda3/lib/python3.8/site-packages/torch/tensor.py:590: RuntimeWarning: Iterating over a tensor might cause the trace to be incorrect. Passing a tensor of different shape won't change the number of iterations executed (and might lead to errors or silently give incorrect results).
  warnings.warn('Iterating over a tensor might cause the trace to be incorrect. '
```
Should we worry? Do you have an idea from where it comes?


2. *error*
```python
from e3nn.util.jit import trace
```
output:
```
ImportError: cannot import name 'compile_mode' from partially initialized module 'e3nn.util.jit' (most likely due to a circular import) (/home/mario/git/e3nn/e3nn/util/jit.py)
```
I already had that error somewhere and I just changed the files without understand what's really going on...

ChangeLog.md
```diff
@@ -5,8 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Recursive TorchScript compiler `e3nn.util.jit`
+- TorchScript support for `TensorProduct` and subclasses, `NormActivation`, `Gate`, `FullyConnectedNet`, and `gate_points_2120.Network`
```
mariogeiger, line 10:
```suggestion
- TorchScript support for `TensorProduct` and subclasses, `NormActivation`, `Gate`, `FullyConnectedNet`, and `gate_points_2101.Network`
```

PR #203, opened 2021-02-17
Fix support for device and dtype

jnooree:
## Description
Major Changes:
- Add some keyword arguments for `torch.device` and `torch.dtype`
- Fix behaviors of some functions that does not work for cuda tensors
- Remove all type-related keyword arguments and internal `.to()` call in module (`torch.nn.Module`) initializers
- Remove all global-scope type modifying code
- Set default dtype of `intertwiners()` and `has_rep_in_rep()` functions to `torch.float64` (`e3nn.math.group.linalg.py`)

Minor Changes:
- Rename `default_dtype.py` to `default_type.py`
- Add `explicit_default_types` fuction for TorchScript compatibility and convenience
- ~Add `torch_default_tensor` and `torch_default_device` context manager-decorator~
- ~Add `add_type_kwargs` decorator for convenience (proposed by @mariogeiger)~
- ~Add utility functions for get/set default device and get default tensor type~
- ~Now `torch_default_dtype` could be used as a decorator~
- ~Now context managers avoid unneccesary operations~
- ~Add `AbstractContextDecoratorManager` class for convenience~

## Motivation and Context
Currently, some functions does not support `torch.Tensor` on non-default `torch.device` and `torch.dtype`.
This pull request add support for `torch.device` and `torch.dtype` for those functions.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:


mariogeiger:
Hi, why adding `device` argument and not `dtype` argument as well?

jnooree:
> Hi, why adding `device` argument and not `dtype` argument as well?

Oh yes, I was a fool. I'll add them right now.

mariogeiger:
idea: can we make a decorator that adds `device` and `dtype` arguments to a function using your new `torch_default_tensor_type` under the hood?

mariogeiger:
Why not removing `torch_default_dtype` ?

jnooree:
> idea: can we make a decorator that adds `device` and `dtype` arguments to a function using your new `torch_default_tensor_type` under the hood?
> 
> ```
> def add_type_args(f):
>     def g(*x, **y, device=None, dtype=None):
>         with torch_default_tensor_type(device, dtype):
>             return f(*x, **y)
>     return g
> ```
> 
> we could even add a `type=None` argument

I think it is possible. One drawback of such wrapper is that it could not preserve the original information of the wrapped function, such as information about arguments and docstrings. `functools.wraps` seems preserving them only if the wrapper passes all of the arguments and keyword arguments.

> Why not removing torch_default_dtype ?

Because it is using the public API directly. `torch_default_tensor_type` is kind of hacky, since `pytorch` does not provide a method equivalent to `torch.get_default_dtype()` for default tensor types.

e3nn/math/normalize_activation.py
```diff
@@ -1,21 +1,27 @@
...
     compute n th moment
     <f(z)^n> for z normal
     """
-    gen = torch.Generator().manual_seed(0)
-    z = torch.randn(1_000_000, generator=gen, dtype=torch.float64)
-    return f(z).pow(n).mean().item()
+    with torch_default_tensor_type(dtype, device):
+        # torch.Generator does not follow the default tensor's device type
+        gen = torch.Generator(device=device).manual_seed(0)
+        z = torch.randn(1_000_000, generator=gen)
```
mariogeiger, line 14: I would like to ensure reproducibility here. Therefore I would like to impose float64 and cpu to generate `z`. Then `z` can be converted

jnooree: Fixed in jnooree/e3nn@a92d418!

e3nn/o3/tensor_product.py
```diff
@@ -156,6 +162,8 @@ def __init__(
             normalization='component',
             internal_weights=None,
             shared_weights=None,
+            dtype=None,
+            device=None,
```
mariogeiger, line 166: Modules usually dont have `dtype/device` arguments, see for instance [torch.nn.Linear](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html). Why do you want it?

jnooree: Because the original code had this line:
https://github.com/e3nn/e3nn/blob/d279900169466a72470b826cf3a96dc75705fadb/e3nn/o3/tensor_product.py#L595
I think calling `.to()` twice is not efficient, but if convention is more important then I'll rollback those changes.

jnooree: @mariogeiger I'll leave them until your decision or comments, just for sure.

mariogeiger:
I'm wondering what is the best option between the following two.
Let say I want to use the following type
```python
dtype = torch.float64
device = 'cuda'
```

Option 1:
Add kargs to `foo` and `Bar` such that we can do
```python
x = foo(..., dtype=dtype, device=device)
f = Bar(..., dtype=dtype, device=device)
f(x)
```

Option 2:
Ensure that `foo` and `Bar` use `torch_get_default_tensor_type` for their types
```python
with torch_default_tensor_type(dtype, device):
    x = foo(...)
    f = Bar(...)
f(x)
```

Option 3:
Make both possible

jnooree:
I think option 3 is the best, for flexibility and convenience. It seems that the `pytorch` modules are just using the default device and dtype without giving option for user, so I think it would be okay to remove the type-related arguments for the modules.

mariogeiger:
@Linux-cpp-lisp @bdice @blondegeek @simonbatzner

mariogeiger:
I think I agree with you @jnooree 
Let just see if the others have an opinion on that

e3nn/o3/s2grid.py
```diff
@@ -265,6 +276,8 @@ class ToS2Grid(torch.nn.Module):
     lmax : int
     res : int, tuple of int
     normalization : {'norm', 'component', 'integral'}
+    dtype : torch.dtype or None, optional
+    device : torch.device or None, optional
```
bdice, line 280: Should these parameters have a comment like in `s2_grid` above?

e3nn/util/default_type.py
```diff
@@ -0,0 +1,85 @@
...
+        def wrapper(*args, dtype=self.dtype, device=self.device, **kwargs):
+            with torch_default_tensor_type(dtype, device):
+                return f(*args, **kwargs)
+
+        if wrapper.__doc__ is not None:
+            if not wrapper.__doc__.endswith("\n"):
+                wrapper.__doc__ += "\n"
+            wrapper.__doc__ += self._DOC_NOTE
+
+        return wrapper
```
bdice, line 72: This class has the bulk of my questions/comments:
1. I'm a little confused why the constructor arguments are needed here. I don't see them being used. You should be able to just supply the parameter defaults as `None` in the `__call__` method.
2. Classes do not need to inherit from `object` anymore, that's an older convention from new/old style classes. You can just write `class MyClass:` and it'll work fine for modern Python versions. Also I don't think you need to call the super constructor for `object`.
3. If you agree that the constructor arguments `dtype` and `device` can be removed, it should be possible to re-write this decorator as a function (not a callable class).
```suggestion
    def __init__(self):
        super().__init__()

    def __call__(self, f):
        @wraps(f)
        def wrapper(*args, dtype=None, device=None, **kwargs):
            with torch_default_tensor_type(dtype, device):
                return f(*args, **kwargs)

        if wrapper.__doc__ is not None:
            if not wrapper.__doc__.endswith("\n"):
                wrapper.__doc__ += "\n"
            wrapper.__doc__ += self._DOC_NOTE

        return wrapper
```

bdice (commented):
Thanks for the tag to review @mariogeiger. I took a look and left a couple comments that may help. Nice work!

Overall I don't have a preference among Options 1-3 stated above. As @jnooree mentioned, the flexibility of Option 3 may be helpful if no other considerations weigh more strongly.

e3nn/o3/s2grid.py
jnooree: I agree, but the other argumetns do not have such long comments so I've followed the style.

e3nn/util/default_type.py
jnooree: > 1. I'm a little confused why the constructor arguments are needed here. I don't see them being used. You should be able to just supply the parameter defaults as `None` in the `__call__` method.
They are not used in this repository yet, but they will make things like this available:

```python3
@add_type_kwargs(dtype=torch.float64)
function_that_will_return_doubletensor_for_default():
    ...

@add_type_kwargs(device="cuda")
function_supposed_to_be_used_on_cuda():
    ...
```
Even these are not being currently used, I think adding such options are better.

> 2. Classes do not need to inherit from `object` anymore, that's an older convention from new/old style classes. You can just write `class MyClass:` and it'll work fine for modern Python versions. Also I don't think you need to call the super constructor for `object`.

You're right, I just added them for the future. Maybe some kind of superclass could be added one day. If this notation is not preferred, then I have no objections removing them.

> 3. If you agree that the constructor arguments `dtype` and `device` can be removed, it should be possible to re-write this decorator as a function (not a callable class).

Actually it was (until jnooree/e3nn@f03a0f2500437740c9e97fe59ccb55d59642256a), but then I decided to add options for flexibility. Please refer the examples above.

mariogeiger: Thanks @bdice for the review. I also favor simple code. more complex code can be added later, I'm not too worried if the signature of a function needs to change in the future, we have a changelog for that.

e3nn/o3/tensor_product.py
mariogeiger: ok for adding those arguments

Linux-cpp-lisp:
Thanks for working on this @jnooree! Very important to get this consistent for running stuff on GPU.

There's a more fundamental issue here than the interface that I'm concerned about: `torch.set_default_dtype` and `torch.set_default_tensor_type` are global state and are **not** thread/parallelism safe. This is an issue (or at least a design decision) in PyTorch — some relevant links are here:

- Open issue on setting global defaults state in thread-safe way: https://github.com/pytorch/pytorch/issues/29720
-  Comment on issues with parallelized PyTorch unit tests when using `torch.set_default_tensor_type`: https://github.com/pytorch/pytorch/issues/27355
- A related PyTorch PR: https://github.com/pytorch/pytorch/pull/27444
- The underlying default dtype code: https://github.com/pytorch/pytorch/blob/69fe10c12788675e8968f9d29530439ac44648bd/c10/core/DefaultDtype.cpp

Hiding this global state inside of seemingly local context managers seems like a recipe for a lot of hard-to-debug concurrency bugs in the future. It's not hard to imagine code that preprocesses data in one thread and runs the model on the other — what happens when they are simultaneously in blocks with contradictory default dtypes/devices? Whose original dtype gets restored? etc.

Unless PyTorch changes their code and offers an official guarantee that default dtype/type are thread-local (including being local to almost-threads like `torch.jit.fork()` and parallel CUDA streams, which could simultaneously call code with different default types), I don't think it's a good idea to use the default dtype/type like this.

As a result, I'm strongly in favor of option 1 (only kwargs) and using those kwargs **explicitly** when functions call `torch.ones`, `torch.eye`, etc. and **not** to set the defaults. I know its not very elegant and leads to some very repetitive `..., device=device, dtype=dtype`, but it is simple and I can't think of another thread-safe approach.

One other important point is that TorchScript does not support `torch.set_default_dtype()` or `torch.set_default_tensor_type()`, and we'd like to be able to call functions like `o3.rand_angles()` from TorchScript models.

Thoughts?

e3nn/o3/tensor_product.py
Linux-cpp-lisp: Was that original line (595) actually necessary? I'm not sure how anything in the `TensorProduct` would, by default, not be of the default dtype.

Is it the Wigner matrices @mariogeiger ?

mariogeiger:
This is a very important concern indeed @Linux-cpp-lisp !

I think we should remove these global state functions then.
@jnooree sorry that it would mean that most of your nice work would need to be removed... so sad

jnooree:
@Linux-cpp-lisp I haven't think about it yet... Thank you for pointing out the potential bugs. I don't want to break the library :D

@mariogeiger It is indeed quite disappointing, but at least I've found most of the "non-obeying" functions so I'm good with it. I'll remove added functions and fallback to option 1.

jnooree:
One question just came to my mind: isn't global interpreter lock exist for that? There would be no more than one thread for any "python code," so there should be no problem for this truely-global variable based approach.

> One other important point is that TorchScript does not support torch.set_default_dtype() or torch.set_default_tensor_type(), and we'd like in general to be able to call functions like o3.rand_angles() from TorchScript models.

If my first question resolves this concurrency issue, then I could fix this problem either. Just removing inheritance could do the job, if I correctly understood `TorchScript` language model.

e3nn/o3/tensor_product.py
jnooree: The original code is long(about 500 lines just for `__init__`), including mostly torchscript generator workarounds. I couldn't figure out the correctness...

Linux-cpp-lisp:
@jnooree like you said, you've found all (and already fixed a few, I think) of the functions that this is relevant to, which is great. Your original solution with the context manager is much nicer than this manual passing around of `device` and `dtype`, but unfortunately the manual passing is the only way PyTorch provides to do this without getting involved with global state.

You're right that the GIL would prevent corruption of the global state itself, but it wouldn't resolve the more fundamental issue — imagine we have two threads both starting with the default tensor type of CPU float32:

| thread 1 | thread 2 |
| ----------- | ----------- |
| old_default_type = CPU,float32 | |
| default_type = GPU,float32 | |
| run some code | old_default_type = GPU,float32|
| default_type = old_default_device | default_type = GPU.float32 |
| | run some code |
| | default_type=old_default_type |

At the end of this example, only thread 1 ran with the correct default type, and the final state is an incorrect default type of `GPU,float32`. (Thread 2 ran with a default type of CPU,float32.) There are infinite variations on this bug. The only solution would be to make default type context blocks mutually exclusive using a [mutex](https://stackoverflow.com/questions/34524/what-is-a-mutex), which would be _very_ inefficient, not to mention impossible in TorchScript.

To your last point, the main issue isn't the inheritance — it's that the `torch.set_default_dtype()` function simply doesn't exist in TorchScript:
```
In [17]: @torch.jit.script
    ...: def f():
    ...:     torch.set_default_dtype(torch.float64)
    ...: 
---------------------------------------------------------------------------
RuntimeError: 
Unknown builtin op: aten::set_default_dtype.
Could not find any similar ops to aten::set_default_dtype. This op may not exist or may not be currently supported in TorchScript.
:
  File "<ipython-input-17-33d4499fc214>", line 3
@torch.jit.script
def f():
    torch.set_default_dtype(torch.float64)
    ~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
```
TorchScript code [doesn't claim the GIL](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html) for efficiency, which I suspect is part of why it doesn't let you touch this global state.

jnooree:
@Linux-cpp-lisp Thanks for detailed examples and multiple corrections. I'll remove all the default modifying behaviors from the code. I hope `pytorch` provide some kind of safer way...

Besides, I still don't get the multithreading environment example, because normally there should be only one thread that is executing the python part, even though the `pytorch` C backend executes multiple of them. Most python parallel codes will use process-based parallelism (by the module `multiprocessing` in standard lib or the well-known external library `joblib`), especially in the field of scientific computing with high cpu/gpu utilization. Multithreading in python is much more useful with networking, GUI, etc., where some threads are just waiting for (relatively) long time. I'm sure that this library isn't targeting such environment, is it?

Linux-cpp-lisp:
You're right that actual threads are pretty rare in Python and not really a concern here — I hadn't thought of that. But one Python thread can still have multiple torchscript (OpenMP, I think) threads in it (`torch.jit.fork`) as well as multiple threads of CUDA execution (CUDA streams on one GPU, or multiple GPU dispatch). I don't know how PyTorch manages this internal state, but there's no official commitment to any consistent behavior (with regards to `torch.set_default_dtype()` and `set_default_tensor_type()`.

For example, does a TorchScript thread always use the default dtype from when it was spawned, or does it fetch it every time it needs it? I have no idea, but because it doesn't claim the GIL it could be executing while the main thread changes the default dtype. Similarly, I have no idea whether, say, [`DataParallel`](https://pytorch.org/docs/stable/generated/torch.nn.DataParallel.html) uses threads or processes.

To be clear: the parallelism is must less of a problem than the TorchScript compatibility. We want as much of the library as possible to be callable from TorchScript, especially numerical helper functions that could plausibly come up inside of a network.

jnooree:
I strongly agree with you that the default modifiers are not (at least, not guaranteed to be) safe for such usage, and TorchScript compatibility is much more critical. I'm removing the context managers.

Linux-cpp-lisp:
Sounds good, thanks @jnooree! Really appreciate all your work on this, it's going to make GPU use much simpler to have this consistent support :+1:

e3nn/o3/tensor_product.py
jnooree: I have removed all type related keyword arguments in modules, as well as `.to()` call inside `__init__()` function. Now the user should call `.to()` method explicitly, if non-default `torch.dtype` and `torch.device` is required.

e3nn/util/default_type.py
jnooree: Removed in jnooree/e3nn@9d0ce0e15ba54a5b0a3bd0e67f72394ec1e9f5fd

e3nn/o3/s2grid.py
jnooree: Removed in jnooree/e3nn@bf83b042eadf08d5a1c8d452eb905ffd6b909e99

mariogeiger:
Thanks for the nice work @jnooree 
And thanks for the nice collaboration @Linux-cpp-lisp @bdice @jnooree

PR #209, opened 2021-02-22
Use unified tensors for weights and wigners in `TensorProduct`

Linux-cpp-lisp:
## Description
Use a single tensor for the `weight`s and `wigner`s inside of `TensorProduct` in a similar way to how multiple irreps are stored in a single tensor.

## Motivation and Context
Avoids dynamic-size list processing in `forward()` and makes `TensorProduct` fully scriptable, removing tracing logic, simplifying future development, and possibly enabling more aggressive TorchScript optimization.

## How Has This Been Tested?
e3nn test suite

## Types of changes

- [X] New feature (non-breaking change which adds or improves functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:

- [X] I have read the [**CONTRIBUTING**](https://github.com/e3nn/e3nn/blob/main/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] I have updated the documentation (if relevant).
- [X] I have added tests that cover my changes (if relevant).
- [X] All new and existing tests passed.
- [X] I have updated the [Changelog](https://github.com/e3nn/e3nn/blob/main/ChangeLog.md).


e3nn/o3/tensor_product.py
```diff
@@ -346,6 +348,11 @@ def indent_for_level(indent_level):
 
             if ins.has_weight:
                 index_w += 1
+                # Extract the weight from the flattened weight tensor
+                line = f"{s}ws_{index_w} = ws[..., {flat_weight_i}:{flat_weight_i + _prod(ins.weight_shape)}].reshape({'' if self.shared_weights else '-1, '}{', '.join(str(d) for d in ins.weight_shape)})\n"
```
mariogeiger, line 352: I think we can simply call it `ws` no need of `ws_{index_w}` ?

e3nn/o3/tensor_product.py
```diff
@@ -365,23 +372,23 @@ def indent_for_level(indent_level):
                     code_right += line
 
                     if ins.connection_mode == 'uvw':
-                        code_out += line_out.format(f"ein('{z}uvw,zu,zv->zw', ws[{index_w}], s1, s2)")
-                        code_right += line_right.format(f"ein('{z}uvw,zv->zuw', ws[{index_w}], s2)")
+                        code_out += line_out.format(f"ein('{z}uvw,zu,zv->zw', ws_{index_w}, s1, s2)")
```
mariogeiger, line 375: then here you just have `ws` and its simpler

mariogeiger (commented):
Nice!

e3nn/o3/tensor_product.py
Linux-cpp-lisp: it would have to be `ws_this` so it doesn't shadow the full weight tensor, but yeah, we could

maybe it's easier to understand with explicit indexes though? No need to figure out where the last line that redefined `ws_this` was. Not sure.

mariogeiger: indeed its cool also as it is now

Linux-cpp-lisp:
Last commit makes the moved `Linear` and `Norm` JITable.

Linux-cpp-lisp:
This should be ready @mariogeiger

PR #218, opened 2021-03-08
[need help] soft_unit_step

mariogeiger:
## Description
I want to implement the function
```
Theta(x) exp(-1/x)
```

where `Theta` is the [Heaviside](https://en.wikipedia.org/wiki/Heaviside_step_function) function.

## Motivation and Context
Useful as an envelope function because it zero for `x <= 0` and converges to one when `x -> infty` while being infinitely differentiable everywhere.
It is often used as a function that is `C^\infty` but non analytical.

## Screenshots
image

## Problem
`pytorch` has trouble differentiating it at `x=0` obviously...

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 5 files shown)


Linux-cpp-lisp:
Maybe just do `torch.exp(-1./x).relu()`?

PR #227, opened 2021-03-18
Update docs for Gate, replace irreps_nonscalars with irreps_gated.

bdice:
## Description
After discussion with Mario and Alby, we decided to improve the docs for `Gate` and replace the argument `irreps_nonscalars` with `irreps_gated` (since the gated irreps could be scalars that are activated by some other scalar).

## Motivation and Context
Improves docs.

## How Has This Been Tested?
I verified that the LaTeX in the docs rendered correctly.

## Screenshots
image

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:


bdice:
I'm going to come back to this tomorrow -- I'm trying one more change, to make the `irreps_in` and `irreps_out` into documented properties, but I might have broken something along the way.

e3nn/nn/gate.py
```diff
@@ -186,15 +207,25 @@ def forward(self, features):
...
+                gated = self.mul(gated, gates)
+                features = torch.cat([scalars, gated], dim=-1)
             else:
                 features = scalars
             return features
+
+    @property
+    def irreps_in(self):
+        """Input representations."""
+        return self._irreps_out
```
mariogeiger, line 226:
```suggestion
        return self._irreps_in
```

e3nn/nn/gate.py
```diff
@@ -122,26 +122,47 @@ def forward(self, x):
...
+    .. math::
+        \left(\bigoplus_i \phi_i(x_i) \right) \oplus \left(\bigoplus_j \phi_j(g_j) y_j \right)
+
+    where :math:`x_i` and :math:`\phi_i` are from ``irreps_scalars`` and
+    ``act_scalars``, and :math:`g_j`, :math:`\phi_j`, and :math:`y_j` are
+    from ``irreps_gates``, ``act_gates``, and ``irreps_gated``.
+
+    The parameters passed in should adhere to two conditions:
+
+    1. ``irreps_scalars.num_irreps == len(act_scalars)``.
```
mariogeiger, line 142:
```suggestion
    1. ``len(irreps_scalars) == len(act_scalars)``.
```
It is not the same!
```
len(Irreps("10x0e + 10xe0")) == 2
Irreps("10x0e + 10xe0").num_irreps == 20
```

e3nn/nn/gate.py
```diff
@@ -122,26 +122,47 @@ def forward(self, x):
...
+    2. ``irreps_gates.num_irreps == len(act_gates) == irreps_gated.num_irreps``.
```
mariogeiger, line 143:
```suggestion
    2. ``len(irreps_gates) == len(act_gates)``
    3. ``irreps_gated.num_irreps == act_gates.num_irreps``.
```

e3nn/nn/gate.py
bdice: Ah, I guessed incorrectly. 🙃 Thanks.

e3nn/nn/gate.py
```diff
@@ -186,15 +207,25 @@ def forward(self, features):
...
+
+    @property
+    def irreps_in(self):
+        """Input representations."""
+        return self._irreps_in
+
+    @property
+    def irreps_out(self):
+        """Output representations (a direct sum of ``irreps_scalars`` and ``irreps_gated``)."""
+        return self._irreps_out
```
Linux-cpp-lisp, line 231: This is definitely good practice, but we also don't do it for any other module in e3nn... trying to think if there's anything this would break

Probably should actually be done for most of the modules? @mariogeiger  Since none of them support changing their irreps after initialization.

mariogeiger: I don’t really see the pro and con to be honest. If you both think it’s a good practice then I will make it everywhere

Linux-cpp-lisp: Pro is that you can't `mod = ...` and later `mod.irreps_in = "apple"`, but that's fairly minor.

I don't think its worth it to make a special effort for other modules, but good to have it here

bdice: @Linux-cpp-lisp I am confident that this won't break anything for properties that are initialized on construction. The primary reason I would choose to do this is that it provides us with a section in the documentation for the public-facing attributes, so that users know of their existence.

Linux-cpp-lisp: Definitely agree it won't break anything.

Didn't occur to me about documentation, I'm used to just putting an `Attributes` section in the class docstring but maybe that doesn't get good results — obviously this is more correct, but does that approach work at all?

bdice: The use of `@property` makes the documentation more automated and it formats itself nicely. 
image
https://e3nn--227.org.readthedocs.build/en/227/api/nn/nn_gate.html#e3nn.nn.gate.Gate

bdice:
@mariogeiger @Linux-cpp-lisp I think this is ready for review/merge. Thanks for the help!

e3nn/nn/gate.py
Linux-cpp-lisp: That looks much nicer :smile:

PR #230, opened 2021-03-20
Create global flake8 configuration.

bdice:
## Description
While working on #227, I noticed that the flake8 configuration used by CI is not stored in the typical place for package/tool configurations (either `setup.cfg` or `.flake8`; [flake8 does not support `pyproject.toml`](https://gitlab.com/pycqa/flake8/-/issues/428)).
I talked with @mariogeiger and we decided to create a package-wide configuration for flake8 that developers can use locally, without needing to copy all the settings from the CI config.
This means that users/developers can simply run `flake8` in the repository root, to check for style issues.

While doing this, I also simplified the flake8 rules from the CI configuration and fixed the small number of issues that appeared. Nothing major has changed -- just normalized style.

## Motivation and Context
One note for context: I have found from working with other packages that it's good to specify a settings like `select = E,F,W,C` for whatever top-level categories of errors that should be checked.
If this is left at the default setting, then users who have flake8 plugins installed will see errors from those plugins even if the code style is compliant for plain flake8 without those plugins installed.
For example, my output of `flake8 --help` shows my plugins:

```
Installed plugins: flake8-bugbear: 20.1.4, mccabe: 0.6.1, naming: 0.11.1, pycodestyle: 2.7.0, pyflakes: 2.3.0, rst-docstrings: 0.0.14
```

Without the setting for `select = E,F,W,C`, I get tons of errors like `N806` from the `naming` plugin, `RST303` errors for unknown directives from `rst-docstrings`, etc., so it's better to just select the basic error classes and ignore user plugins.

## How Has This Been Tested?
I ran flake8 locally until the number of included/ignored rules had been reduced to a reasonably minimal set. :)

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation improvement (updates to user guides, docstrings, or developer docs)

## Checklist:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,10 +28,8 @@ jobs:
         pip install flake8
     - name: Lint with flake8
       run: |
-        # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82,E251,F811,E302,E261,E231,F401,F841,W293,W291,W605,W292,W391,E128,E305,W503,W504,E303,E125,E121,E123,E203,F541,E713 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 e3nn --count --exit-zero --statistics --ignore=E741,E743,E501,C901
+        # Stop the build if there are style issues, Python syntax errors, or undefined names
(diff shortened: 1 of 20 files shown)


mariogeiger:
Very cool !

PR #251, opened 2021-04-09
Avoids re-entrant file opening for Windows

mister-bailey:
As per Mario Geiger and Alby M's suggestion, this uses a fixed filename for the temporary code, but in a temporary directory.


bdice (approved):
This looks correct from the Slack conversation we had.

PR #313, opened 2021-11-04
Compile minimum stuff

mariogeiger:
- split codegen in two: `left_right` and `right`
- add flags `compile_left_right` and `compile_right` to `TensorProduct`
- set default `optimize_einsums=False`
- add new global flag `jit_script_fx=False` that by default turns of the `torch.jit.script` of `fx` code

### DIFF
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -10,23 +10,18 @@ _OPT_DEFAULTS: Dict[str, bool] = dict(
 )
 
 
-def set_optimization_defaults(
-    specialized_code: bool = True,
-    optimize_einsums: bool = False,
-    jit_script_fx: bool = False,
-) -> None:
+def set_optimization_defaults(**kwargs) -> None:
     r"""Globally set the default optimization settings.
 
     Parameters
     ----------
-    specialized_code : bool, default True
-        Whether to use specialized code for (combinations of) irreps for which it exists.
-    optimize_einsums : bool, default True
-        Whether to use ``opt_einsum_fx``.
+    **kwargs
+        Keyword arguments to set the default optimization settings.
     """
-    _OPT_DEFAULTS['specialized_code'] = specialized_code
-    _OPT_DEFAULTS['optimize_einsums'] = optimize_einsums
-    _OPT_DEFAULTS['jit_script_fx'] = jit_script_fx
+    for k, v in kwargs.items():
+        if k not in _OPT_DEFAULTS:
+            raise ValueError(f"Unknown optimization option: {k}")
+        _OPT_DEFAULTS[k] = v
 
 
 def get_optimization_defaults() -> Dict[str, bool]:
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -5,13 +5,15 @@ from typing import Dict
 
 _OPT_DEFAULTS: Dict[str, bool] = dict(
     specialized_code=True,
-    optimize_einsums=True
+    optimize_einsums=False,
+    jit_script_fx=False,
 )
 
 
 def set_optimization_defaults(
     specialized_code: bool = True,
-    optimize_einsums: bool = True
+    optimize_einsums: bool = False,
+    jit_script_fx: bool = False,
 ) -> None:
     r"""Globally set the default optimization settings.
 
@@ -24,6 +26,7 @@ def set_optimization_defaults(
     """
     _OPT_DEFAULTS['specialized_code'] = specialized_code
     _OPT_DEFAULTS['optimize_einsums'] = optimize_einsums
+    _OPT_DEFAULTS['jit_script_fx'] = jit_script_fx
 
 
 def get_optimization_defaults() -> Dict[str, bool]:
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -6,7 +6,7 @@ from typing import Dict
 _OPT_DEFAULTS: Dict[str, bool] = dict(
     specialized_code=True,
     optimize_einsums=False,
-    jit_script_fx=False,
+    jit_script_fx=True,
 )
 
 
--- a/e3nn/nn/models/v2103/voxel_convolution.py
+++ b/e3nn/nn/models/v2103/voxel_convolution.py
@@ -60,7 +60,12 @@ class Convolution(torch.nn.Module):
         sh = o3.spherical_harmonics(self.irreps_sh, lattice, True, 'component')  # [x, y, z, irreps_sh.dim]
         self.register_buffer('sh', sh)
 
-        self.tp = FullyConnectedTensorProduct(self.irreps_in, self.irreps_sh, self.irreps_out, shared_weights=False)
+        self.tp = FullyConnectedTensorProduct(
+            self.irreps_in, self.irreps_sh, self.irreps_out,
+            shared_weights=False,
+            compile_left_right=False,
+            compile_right=False,
+        )
 
         self.weight = torch.nn.Parameter(torch.randn(self.num_rbfs, self.tp.weight_numel))
 
--- a/e3nn/nn/models/v2103/voxel_convolution.py
+++ b/e3nn/nn/models/v2103/voxel_convolution.py
@@ -64,7 +64,7 @@ class Convolution(torch.nn.Module):
             self.irreps_in, self.irreps_sh, self.irreps_out,
             shared_weights=False,
             compile_left_right=False,
-            compile_right=False,
+            compile_right=True,
         )
 
         self.weight = torch.nn.Parameter(torch.randn(self.num_rbfs, self.tp.weight_numel))
--- a/e3nn/o3/_linear.py
+++ b/e3nn/o3/_linear.py
@@ -1,6 +1,6 @@
 from typing import List, NamedTuple, Optional, Tuple, Union
 
-from opt_einsum_fx import jitable, optimize_einsums_full
+from opt_einsum_fx import optimize_einsums_full
 import torch
 from torch import fx
 
@@ -489,6 +489,6 @@ def _codegen_linear(
                 bias_numel,
             ),
         )
-        graphmod_out = jitable(optimize_einsums_full(graphmod_out, example_inputs))
+        graphmod_out = optimize_einsums_full(graphmod_out, example_inputs)
 
     return graphmod_out, weight_numel, bias_numel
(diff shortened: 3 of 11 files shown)


e3nn/__init__.py
```diff
@@ -5,13 +5,15 @@
 
 _OPT_DEFAULTS: Dict[str, bool] = dict(
     specialized_code=True,
-    optimize_einsums=True
+    optimize_einsums=False,
+    jit_script_fx=False,
```
Linux-cpp-lisp, line 9: Why `False` as a default? This will bring back that weird mid-training stack overflow bug that we only avoid by scripting FX modules.

Nice to support it both ways, but I don't think a default should enable a known bug?

e3nn/__init__.py
```diff
@@ -24,6 +26,7 @@ def set_optimization_defaults(
     """
     _OPT_DEFAULTS['specialized_code'] = specialized_code
     _OPT_DEFAULTS['optimize_einsums'] = optimize_einsums
+    _OPT_DEFAULTS['jit_script_fx'] = jit_script_fx
```
Linux-cpp-lisp, line 29: My fault, but this method makes no sense... it should only be updating keys that are actually passed as parameters, and shouldn't touch others.

Something with `**kwargs`, check that keys are a subset of those in _OPT_DEFAULTS, and then `_OPT_DEFAULTS.update(kwargs)`...?

e3nn/nn/models/v2103/voxel_convolution.py
```diff
@@ -60,7 +60,12 @@ def __init__(self, irreps_in, irreps_out, irreps_sh, size, steps=(1, 1, 1), **kw
         sh = o3.spherical_harmonics(self.irreps_sh, lattice, True, 'component')  # [x, y, z, irreps_sh.dim]
         self.register_buffer('sh', sh)
 
-        self.tp = FullyConnectedTensorProduct(self.irreps_in, self.irreps_sh, self.irreps_out, shared_weights=False)
+        self.tp = FullyConnectedTensorProduct(
+            self.irreps_in, self.irreps_sh, self.irreps_out,
+            shared_weights=False,
+            compile_left_right=False,
```
Linux-cpp-lisp, line 66: Can you explain the name of this flag? Why `left_right`? Does it compile both `forward` and `right`? In which case, why isn't there a way to only compile `forward`?

e3nn/o3/_linear.py
```diff
@@ -489,6 +489,6 @@ def _codegen_linear(
                 bias_numel,
             ),
         )
-        graphmod_out = jitable(optimize_einsums_full(graphmod_out, example_inputs))
+        graphmod_out = optimize_einsums_full(graphmod_out, example_inputs)
```
Linux-cpp-lisp, line 492: why remove `jitable`? At least, you still need to call it if `jit_script_fx` is true.

I know they un-broke some stuff with the interface compatability between python and torchscript, but is it all fixed? do we know that? and by what version? since we still support 1.8

e3nn/o3/_tensor_product/_codegen.py
```diff
@@ -20,7 +20,7 @@ def _sum_tensors(xs: List[torch.Tensor], shape: torch.Size, like: torch.Tensor):
     return like.new_zeros(shape)
 
 
-def codegen_tensor_product(
+def codegen_tensor_product_left_right(
```
Linux-cpp-lisp, line 23: same question about `left_right` terminology

e3nn/o3/_tensor_product/_tensor_product.py
```diff
@@ -191,6 +197,8 @@ def __init__(
         path_normalization: str = 'element',
         internal_weights: Optional[bool] = None,
         shared_weights: Optional[bool] = None,
+        compile_left_right: bool = True,
+        compile_right: bool = False,
```
Linux-cpp-lisp, line 201: :+1:

e3nn/o3/_tensor_product/_tensor_product.py
```diff
@@ -318,19 +326,45 @@ def num_elements(ins):
...
+                "_compiled_main_right": graphmod_right
+            })
+        elif graphmod_left_right is not None:
+            self._codegen_register({
+                "_compiled_main_left_right": graphmod_left_right,
+            })
+        elif graphmod_right is not None:
+            self._codegen_register({
+                "_compiled_main_right": graphmod_right
+            })
```
Linux-cpp-lisp, line 367: see comment below

e3nn/o3/_tensor_product/_tensor_product.py
```diff
@@ -440,6 +474,9 @@ def right(self, y, weight: Optional[torch.Tensor] = None):
         """
         assert y.shape[-1] == self._in2_dim, "Incorrect last dimension for y"
 
+        if not hasattr(self, '_compiled_main_right'):
+            raise ValueError("`right` method is not compiled, set `compile_right` to True when creating the TensorProduct")
```
Linux-cpp-lisp, line 478: not sure that this is torchscript compatable

better would be to generate a tiny `GraphModule` that just throws a useful error for `graphmod_right` if `not compile_right`. This should do it:

```python
graph = fx.Graph()
graph.placeholder("...")  # give it relevant arguments so it doesn't die on wrong number of args
graph.call_function(torch._assert, args=(False, "`right` method is not compiled, set `compile_right` to True when creating the TensorProduct")
```

Linux-cpp-lisp: this also eliminates the ugly conditional in init for calling `_codegen_register`

Linux-cpp-lisp: it also handles the right-only case correctly (this check is currently missing from forward)

e3nn/util/codegen/_mixin.py
```diff
@@ -31,10 +33,17 @@ def _codegen_register(
             self.__codegen__ = []
         self.__codegen__.extend(funcs.keys())
 
+        opt_defaults = e3nn.get_optimization_defaults()
```
Linux-cpp-lisp, line 36: see discussing in Slack; TensorProduct should really take a `optimization_options` dict and should propagate it through to CodeGenMixin so that it is possible to override this on an object-by-object basis

low importance tho

e3nn/util/codegen/_mixin.py
```diff
@@ -31,10 +33,17 @@ def _codegen_register(
...
+
         for fname, graphmod in funcs.items():
             assert isinstance(graphmod, fx.GraphModule)
-            scriptmod = torch.jit.script(graphmod)
-            assert isinstance(scriptmod, torch.jit.ScriptModule)
+
+            if opt_defaults['jit_script_fx']:
+                scriptmod = torch.jit.script(jitable(graphmod))
```
Linux-cpp-lisp, line 42: aha you call jitable here, good

e3nn/o3/_linear.py
Linux-cpp-lisp: nevermind I see it at the bottom

e3nn/util/codegen/_mixin.py
```diff
@@ -96,6 +107,8 @@ def __setstate__(self, d):
                 # Load the TorchScript IR buffer
                 buffer = io.BytesIO(buffer)
                 smod = torch.jit.load(buffer)
+                if isinstance(smod, fx.GraphModule):
+                    smod = torch.jit.script(jitable(smod))
```
Linux-cpp-lisp, line 111: since we're in `__setstate__` it can't be a GraphMod, it has to be a scriptmodule... since thats the only thing we can save

Linux-cpp-lisp:
```suggestion
```

e3nn/util/codegen/_mixin.py
```diff
@@ -62,6 +71,8 @@ def __getstate__(self):
             for fname in self.__codegen__:
                 # Get the module
                 smod = getattr(self, fname)
+                if isinstance(smod, fx.GraphModule):
+                    smod = torch.jit.script(jitable(smod))
```
Linux-cpp-lisp, line 75: so this is correct in the sense that it does need to be a scriptmodule to save it... but raises the question again of why you need a no-compile option in the first place?

Linux-cpp-lisp: (in theory I think you _can_ now pickle a GraphModule; I don't fully remember why we abandoned that but I'm sure there was a good reason... probably has something to do with how broken that was on 1.8)

e3nn/nn/models/v2103/voxel_convolution.py
mariogeiger: left_right means that you tensor product with both left and right hand side. 

forward == left_right
.right == right

e3nn/o3/_tensor_product/_codegen.py
mariogeiger: same answer

e3nn/__init__.py
mariogeiger: done

e3nn/__init__.py
mariogeiger: Because it can take up to 10s to instantiate a `TensorProdcut` otherwise...

e3nn/o3/_tensor_product/_tensor_product.py
mariogeiger: done

e3nn/util/codegen/_mixin.py
mariogeiger: Also agree but I think we can do that in a separate PR

e3nn/__init__.py
Linux-cpp-lisp: Having a known bug that will only show up later in training by default just feels like a way to make a lot of people very upset 😄

e3nn/nn/models/v2103/voxel_convolution.py
Linux-cpp-lisp: 👍

e3nn/util/codegen/_mixin.py
Linux-cpp-lisp: 👍

e3nn/__init__.py
mariogeiger: I will measure the different execution times and we can descide

mariogeiger:
https://slides.com/mariogeiger/e3nn-optim-options/

PR #315, opened 2021-11-16
Faster CI tests

Linux-cpp-lisp:
Don't install unneeded torchvision and torchaudio

No reason to waste CI time installing dependencies that are never used by e3nn.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install wheel
-        pip install torch==${TORCH} -f https://download.pytorch.org/whl/cpu/torch_stable.html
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install wheel
-        pip install torch==${TORCH} torchvision torchaudio -f https://download.pytorch.org/whl/cpu/torch_stable.html
+        pip install torch==${TORCH} -f https://download.pytorch.org/whl/cpu/torch_stable.html
```
Linux-cpp-lisp, line 42:
```suggestion
        pip install torch==${TORCH} torchvision -f https://download.pytorch.org/whl/cpu/torch_stable.html
```

PR #320, opened 2021-11-23
Remove CartesianTensor._rtp

mariogeiger:
```python
irreps = CartesianTensor("ij=-ji")

rtp = irreps.reduced_tensor_products()
# rtp is a nn.Module

irreps.from_cartesian(data, rtp=rtp)  # use rtp

irreps.from_cartesian(data)  # create a new ReducedTensorProducts at each call
```

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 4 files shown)


blondegeek:
Can the user still fix which rtp is used to convert the data? There are
users who need to do this because of the degeneracy of the change of basis
matrix and they may have existing matrices.

mariogeiger:
"sill"? I don't think the user can do that right now.

I just made `ReducedTensorProducts` pickle-compatible using `CodeGenMixin`.
The user will be able to do that:

```python
irreps = CartesianTensor("ij=-ji")
torch.save(irreps.reduced_tensor_products(), "rtp")

# 6 month later...
irreps = CartesianTensor("ij=-ji")
rtp = torch.load("rtp")
irreps.from_cartesian(data, rtp=rtp)  # use rtp
```

blondegeek:
Great! Thanks!

mjwen:
Great solution, thanks!

PR #321, opened 2021-11-29
Added `from e3nn.o3.irrep import lXY ` syntax sugar

dmadisetti:
In your MRS talk, and going through the examples. I thought I would suggest this quick feature similar to `sympy`'s var: https://docs.sympy.org/latest/modules/core.html#var

## Description

Here, we can generate an atom by import opposed to string creation. Example:

old:
```python
from e3nn.o3 import  Irrep
l3o = Irrep("3o")
```

new:
```python
from e3nn.o3.irrep import l3o
```

## Motivation and Context
This syntax is a little less bulky, but is totally not needed. Feel free to reject this change.

## How Has This Been Tested?
Appended to test in irreps.creation. Let me know if you'd like more. Tested in colab, but hoping your CI will pick it up.

## Checklist:
- ? My code follows the code style of this project.
- ? I have updated the documentation (if relevant).
- [x] The modified code is cuda compatible (github tests don't test cuda) (if relevant).


mariogeiger (approved):

dmadisetti:
Thanks for the flake!

dmadisetti:
I'll take a look at this. Sorry about it breaking, that's a bit embarrassing. I'll check 3.9, 3.7 is what I tested on

dmadisetti:
Solve! I think. `pytest` does some weird module attribute spoofing stuff. Using `__getattribute__` over `__getattr__` allows for pytest to fallback (and is better anyway).

tests/o3/irreps_test.py
```diff
@@ -19,6 +19,8 @@ def test_creation():
     o3.Irreps([(16, "3e"), '1e'])
     o3.Irreps([(16, "3e"), '1e', (256, (1, -1))])
 
+    assert o3.irreps.l3e == ir
```
mariogeiger, line 22: Is it `o3.irreps` or `o3.irrep` ? `o3.irrep` looks better to me

dmadisetti:
Great, I'm also going to make a PR suggesting one runs `flake8 . --count --show-source --statistics` since this has tripped me up a few times.

dmadisetti:
Ok, I figure out testing, and I figured out flake.

This is still broken, so I'm turning this to draft. I will fix it, but it's not as simple as I thought with pytest

dmadisetti:
Ok tests work locally. Sorry for taking so long on such a silly and small change

tests/o3/irreps_test.py
dmadisetti: irrep it is!

mariogeiger:
Let me give up the support of python 3.6

PR #341, opened 2022-04-09
cg from qutip

mariogeiger:
Use the [change of basis from real to complex](https://en.wikipedia.org/wiki/Spherical_harmonics#Real_form) given by the wikipedia page of the spherical harmonics to compute the clebsch-gordan coefficient of so3 from the [clebsch-gordan coefficient of su2 given by the qutip library](https://qutip.org/docs/3.1.0/apidoc/functions.html#qutip.utilities.clebsch).

This imply a change of sign for the clebsch-gordan coefficients for some triplet.
This is due to the fact that in main, the way to fix the sign of the cg was arbitrary. Using qutip cg fixes this arbitrariness.
```
l1,l2,l3=(1,2,2): main == -qutip
l1,l2,l3=(1,3,3): main == -qutip
l1,l2,l3=(1,4,4): main == -qutip
l1,l2,l3=(1,5,5): main == -qutip
l1,l2,l3=(1,6,6): main == -qutip
l1,l2,l3=(1,7,7): main == -qutip
l1,l2,l3=(1,8,8): main == -qutip
l1,l2,l3=(2,1,2): main == -qutip
l1,l2,l3=(2,2,1): main == -qutip
l1,l2,l3=(3,1,3): main == -qutip
l1,l2,l3=(3,3,1): main == -qutip
l1,l2,l3=(3,4,6): main == -qutip
l1,l2,l3=(3,5,7): main == -qutip
l1,l2,l3=(3,6,4): main == -qutip
l1,l2,l3=(3,6,8): main == -qutip
l1,l2,l3=(3,7,5): main == -qutip
l1,l2,l3=(3,8,6): main == -qutip
l1,l2,l3=(4,1,4): main == -qutip
l1,l2,l3=(4,3,6): main == -qutip
l1,l2,l3=(4,4,1): main == -qutip
l1,l2,l3=(4,4,5): main == -qutip
l1,l2,l3=(4,5,4): main == -qutip
l1,l2,l3=(4,5,6): main == -qutip
l1,l2,l3=(4,6,3): main == -qutip
l1,l2,l3=(4,6,5): main == -qutip
l1,l2,l3=(4,6,7): main == -qutip
l1,l2,l3=(4,7,6): main == -qutip
l1,l2,l3=(4,7,8): main == -qutip
l1,l2,l3=(4,8,7): main == -qutip
l1,l2,l3=(5,1,5): main == -qutip
l1,l2,l3=(5,3,7): main == -qutip
l1,l2,l3=(5,4,4): main == -qutip
l1,l2,l3=(5,4,6): main == -qutip
l1,l2,l3=(5,5,1): main == -qutip
l1,l2,l3=(5,5,5): main == -qutip
l1,l2,l3=(5,6,4): main == -qutip
l1,l2,l3=(5,6,6): main == -qutip
l1,l2,l3=(5,7,3): main == -qutip
l1,l2,l3=(5,7,7): main == -qutip
l1,l2,l3=(5,8,8): main == -qutip
l1,l2,l3=(6,1,6): main == -qutip
l1,l2,l3=(6,3,4): main == -qutip
l1,l2,l3=(6,3,8): main == -qutip
l1,l2,l3=(6,4,3): main == -qutip
l1,l2,l3=(6,4,5): main == -qutip
l1,l2,l3=(6,4,7): main == -qutip
l1,l2,l3=(6,5,4): main == -qutip
l1,l2,l3=(6,5,6): main == -qutip
l1,l2,l3=(6,6,1): main == -qutip
l1,l2,l3=(6,6,5): main == -qutip
l1,l2,l3=(6,7,4): main == -qutip
l1,l2,l3=(6,8,3): main == -qutip
l1,l2,l3=(7,1,7): main == -qutip
l1,l2,l3=(7,3,5): main == -qutip
l1,l2,l3=(7,4,6): main == -qutip
```

### DIFF
--- a/e3nn/o3/_wigner.py
+++ b/e3nn/o3/_wigner.py
@@ -84,9 +84,9 @@ def wigner_D(l, alpha, beta, gamma):
         tensor :math:`D^l(\alpha, \beta, \gamma)` of shape :math:`(2l+1, 2l+1)`
     """
     alpha, beta, gamma = torch.broadcast_tensors(alpha, beta, gamma)
-    alpha = alpha[..., None, None]
-    beta = beta[..., None, None]
-    gamma = gamma[..., None, None]
+    alpha = alpha[..., None, None] % (2 * torch.pi)
+    beta = beta[..., None, None] % (2 * torch.pi)
+    gamma = gamma[..., None, None] % (2 * torch.pi)
     X = so3_generators(l)
     return torch.matrix_exp(alpha * X[1]) @ torch.matrix_exp(beta * X[0]) @ torch.matrix_exp(gamma * X[1])
 
--- a/e3nn/o3/_wigner.py
+++ b/e3nn/o3/_wigner.py
@@ -1,37 +1,54 @@
 r"""Core functions of :math:`SO(3)`
 """
-import os
-
 import torch
 
-from e3nn import o3
 from e3nn.util import explicit_default_types
 
-_Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
-# _Jd is a list of tensors of shape (2l+1, 2l+1)
-# _W3j_flat is a flatten version of W3j symbols
-# _W3j_indices is a dict from (l1, l2, l3) -> slice(i, j) to index the flat tensor
-# only l1 <= l2 <= l3 are stored
+
+def su2_generators(j) -> torch.Tensor:
+    m = torch.arange(-j, j)
+    raising = torch.diag(-torch.sqrt(j * (j + 1) - m * (m + 1)), diagonal=-1)
+
+    m = torch.arange(-j + 1, j + 1)
+    lowering = torch.diag(torch.sqrt(j * (j + 1) - m * (m - 1)), diagonal=1)
+
+    m = torch.arange(-j, j + 1)
+    return torch.stack([
+        0.5 * (raising + lowering),  # x (usually)
+        torch.diag(1j * m),  # z (usually)
+        -0.5j * (raising - lowering),  # -y (usually)
+    ], dim=0)
 
 
-def _z_rot_mat(angle, l):
-    r"""
-    Create the matrix representation of a z-axis rotation by the given angle,
-    in the irrep l of dimension 2 * l + 1, in the basis of real centered
-    spherical harmonics (RC basis in rep_bases.py).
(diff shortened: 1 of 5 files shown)


e3nn/o3/_wigner.py
```diff
@@ -118,76 +130,139 @@ def wigner_3j(l1, l2, l3, flat_src=_W3j_flat, dtype=None, device=None):
...
+    j3, m3 = idx3
+
+    if m3 != m1 + m2:
+        return 0
+    vmin = int(max([-j1 + j2 + m3, -j1 + m1, 0]))
+    vmax = int(min([j2 + j3 + m1, j3 - j1 + j2, j3 + m3]))
+
+    def f(n):
+        assert n == round(n)
+        return factorial(round(n))
```
muhrin, line 252: I've found it to be a huge speedup to cache these factorials, in my code I use something like:

```
@functools.lru_cache(maxsize=None)
def factorial(n: int) -> int:
    return math.factorial(n)
```

`maxsize` might need to be set if you think there's a chance users could blow through their RAM.

mariogeiger: I am very surprised that factorial are slow enough to require cache!! 
I will give it a try

muhrin:
Very nice :) Btw, do these include the Condon-Shortley phase?

mariogeiger:
> Very nice :) Btw, do these include the Condon-Shortley phase?

I don't know, let me check

e3nn/o3/_wigner.py
muhrin: Yes, it might be less impactful on typical e3nn workflows.  For my case, I was doing many thousands (or more) of transformations between geometric and spherical tensors so the factorials were a non-trivial component of my runtime.

mariogeiger:
> Very nice :) Btw, do these include the Condon-Shortley phase?

The change of basis I took from wikipedia says that it has the Condon-Shortley phase but I don't really understand where it plays a role

muhrin:
I think it does by having the `(-1)^m` in the CoB matrices

e3nn/o3/_wigner.py
muhrin: Maybe I'm wrong:

```
In [9]: %timeit math.factorial(1024)
48.6 µs ± 67.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Maybe my speed-up came from caching binomial coefficients that I also have to calculate...

PR #408, opened 2023-07-08
Combine` xx = torch.einsum(..., x1, x2)` into `result = torch.einsum(...)`

mitkotak:
- [ ] Need to find a better alternative for
```python
xx = torch.einsum("zui,zvj->zuvij", x1, x2)[:, i[0], i[1]]
```

### DIFF
--- a/e3nn/o3/_tensor_product/_codegen.py
+++ b/e3nn/o3/_tensor_product/_codegen.py
@@ -98,9 +98,6 @@ def codegen_tensor_product_left_right(
     # The einsum string index to prepend to the weights if the weights are not shared and have a batch dimension
     z = "" if shared_weights else "z"
 
-    # Cache of input irrep pairs whose outer products (xx) have already been computed
-    xx_dict = dict()
-
     # Current index in the flat weight tensor
     flat_weight_index = 0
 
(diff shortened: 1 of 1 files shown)


mariogeiger:
One way could be to add a one-hot matrix in the einsume instead of using indexing.

Did you tested that this changes makes the code faster?

Linux-cpp-lisp:
I'm pretty sure `opt_einsum_fx` was already doing this optimization with `_fuse.py`, but maybe not; should depend on the reuse pattern of `xx`:

https://github.com/Linux-cpp-lisp/opt_einsum_fx/blob/main/opt_einsum_fx/_fuse.py

mariogeiger:
Oh cool! I forgot `opt_einsum_fx` had a fusing operation

Linux-cpp-lisp:
Yeah, it does it only in the case where the first einsum is used only by a single other einsum though: 
https://github.com/Linux-cpp-lisp/opt_einsum_fx/blob/main/opt_einsum_fx/_fuse.py#L79-L83
The assumption was that the user knew what they were doing if they used it in two places and then the precomputing should happen once. But that could of course be made an option.

PR #436, opened 2024-06-13
`o3.experimental.FullTensorProductv2` for `torch.compile(...., fullgraph=True)`

mitkotak:
@mariogeiger @Linux-cpp-lisp Would appreciate a review. Also let me know what other tests I can add to `tests/o3/experimental/test_fulltp.py`

Linear coming in next !

### DIFF
--- a/e3nn/o3/__init__.py
+++ b/e3nn/o3/__init__.py
@@ -57,7 +57,7 @@ from ._s2grid import (
 from ._so3grid import SO3Grid
 from ._linear import Linear
 from ._norm import Norm
-
+from .experimental import *
 
 __all__ = [
     "rand_matrix",
--- a/e3nn/o3/__init__.py
+++ b/e3nn/o3/__init__.py
@@ -38,6 +38,8 @@ from ._tensor_product import (
     FullTensorProduct,
     TensorSquare,
 )
+from .experimental import FullTensorProductv2
+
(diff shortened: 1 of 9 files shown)


mitkotak:
Here's what the numbers look like on RTX A5500

timing

e3nn/o3/__init__.py
```diff
@@ -57,7 +57,7 @@
 from ._so3grid import SO3Grid
 from ._linear import Linear
 from ._norm import Norm
-
+from .experimental import *
```
mariogeiger, line 60: Please import one by one and add in the all list below

e3nn/util/datatypes.py
```diff
@@ -0,0 +1,13 @@
+from typing import NamedTuple, Optional
+
+
+class Path(NamedTuple):
```
mariogeiger, line 4: I'd call this a Chunk maybe

e3nn/util/datatypes.py
```diff
@@ -0,0 +1,13 @@
...
+    mul: int
+    dim: int
+    slice: Optional[slice] = None
+
+
+class Paths(NamedTuple):
```
mariogeiger, line 10: And this would be a Path

mariogeiger (approved):
Hello Mit, looks good👌

mitkotak:
@mariogeiger Thanks for the review ! If all looks good then can you merge ?

PR #437, opened 2024-06-22
Fix TorchScript warning "doesn't support instance-level annotations"

Linux-cpp-lisp:
## Description
In recent PyTorch versions, this warning has started coming up a lot:
```
UserWarning: The TorchScript type system doesn't support instance-level annotations on empty non-base types in `__init__`. Instead, either 1) use a type annotation in the class body, or 2) wrap the type in `torch.jit.Attribute`.
```

## How Has This Been Tested?
All unit tests pass locally, with GPU.

## Checklist:

- [ X ] I have read the [**CONTRIBUTING**](https://github.com/e3nn/e3nn/blob/main/.github/CONTRIBUTING.md) document.
- [ X ] My code follows the code style of this project.
- [ X ] I have updated the documentation (if relevant).
- [ X ] I have added tests that cover my changes (if relevant).
- [ X ] The modified code is cuda compatible (github tests don't test cuda) (if relevant).
- [ ] I have updated the [CHANGELOG](https://github.com/e3nn/e3nn/blob/main/.github/CHANGELOG.md).

PR #438, opened 2024-06-23
Fixing broken tests in #437

mitkotak:
- [x] Explicitly runs `tests` and `examples` instead of running it from source (similar to [e3nn-jax](https://github.com/e3nn/e3nn-jax/blob/a2a81ab451b9cd597d7be27b3e1faba79457475d/.github/workflows/tests.yml#L42))

- [x] Ignores `experimental` folder in `pytest` workflow

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,7 +53,7 @@ jobs:
         pip install coveralls
     - name: Test with pytest
       run: |
-        coverage run --source=e3nn -m pytest --doctest-modules --ignore=docs/ .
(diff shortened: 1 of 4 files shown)


mitkotak:
Will resolve the docs issue in a separate PR. Seems to be coming from one of the [notebooks](https://readthedocs.org/projects/e3nn/builds/24787661/).

@mariogeiger  Can I get a review and merge so that testing is back on track ?

Linux-cpp-lisp:
Should the tests just use the latest torch `2.*`, rather than being pegged to 2.0?  `numpy` 2.0 seems to work ok in our tests when torch 2.* is used.

mariogeiger (approved):

mariogeiger:
Thanks @mitkotak !

PR #446, opened 2024-07-10
[WIP] Linear pt2

mitkotak:
Starting draft for Linear PT2

Benchmark on RTX A5500

timing


JoshRackers:
Would love to use this! Is there any thing I can do to help get this merged?

mitkotak:
Hey nice to see you (Atomic Architects say Hi)

Turns out we don't need all of this. Just doing

```python
e3nn.set_optimization_defaults(jit_script_fx=False)
torch.compile(..., fullgraph=True)
```

makes it work with all of the existing code ! And the gains more or less remain the same

linear_pt2


The snippet is all you need to make it work but I am also testing this out and providing some wrappers [here](https://github.com/e3nn/e3nn/pull/455). It's mostly ready to land, just some legwork in powering through test fixes. Would appreciate some help there if you are interested

JoshRackers:
Hi Mit- nice to see you too! Tell everyone in the lab I said hi!

That is very funny (and great!) that this work so well just out of the box like this. We will give this a shot and report back. We are mainly interested in accelerating the various tensor product classes, as I suspect most folks are.

mitkotak:
Yeah haha. Actually full credit to Alby for figuring this one out.

PR #453, opened 2024-07-26
v0.5.2

Linux-cpp-lisp:


mitkotak:
@Linux-cpp-lisp Are we drafting a release or is Mario doing it ?

Linux-cpp-lisp:
Release can just be the excerpt from the changelog, no?

PR #458, opened 2024-08-05
Fixing release workflow

mitkotak:
## Problem

The release workflow [did not run](https://github.com/e3nn/e3nn/actions/runs/10120348432/job/27989989135) for the latest release

## Solution

Doing what NequIP does and explicitly downloading all dependencies should hopefully fix the issue.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,12 +17,13 @@ jobs:
         python-version: 3.8
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip build
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
     - name: Build and publish
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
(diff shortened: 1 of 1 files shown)


mitkotak:
@Linux-cpp-lisp @mariogeiger

mariogeiger:
It's sus to see a setup.py when we switched to pyproject.toml

mariogeiger:
I would recommend switching to https://hatch.pypa.io/latest/ if that can help

mitkotak:
@mariogeiger Upgraded to hatch. Was able to get to this locally but I am wondering how to test the workflow

```
adding 'e3nn-0.5.2.dist-info/RECORD'
removing build/bdist.linux-x86_64/wheel
Successfully built e3nn-0.5.2.tar.gz and e3nn-0.5.2-py3-none-any.whl
(.venv) mkotak@radish:~/atomic_architects/lib/e3nn$ 
```

mariogeiger:
I don't know how to test other than making a release and look if it works. 

On another note I would suggest to use hatch for pyproject.toml as well

PR #483, opened 2024-10-31
[WIP] Fixed CPU loading of a GPU tensor product

### LINKED ISSUE #461: 🐛 [BUG] Can't map location of tensor product to cpu
**Describe the bug**
Same as described in #460 

If I save a tensor product that has been moved to the gpu, I can't load it if there's no gpu available.

**To Reproduce**

Store two checkpoint files, one with the tensor product in the CPU and one with the tensor product on the gpu:
```python
from e3nn import o3
import torch

tp = o3.FullyConnectedTensorProduct("40x0e+40x1o+40x2e", "1x0e+1x1o+1x2e", "120x0e+160x1o+160x2e")

torch.save(tp, "test_cpu.ckpt")

tp = tp.to("cuda")

torch.save(tp, "test_gpu.ckpt")
```

Then, in an environment with no cuda available, try to load them both mapping to cpu:

```python
import torch

cpt = torch.load("test_cpu.ckpt", map_location="cpu")

print("LOADED CPU")

cpt = torch.load("test_gpu.ckpt", map_location="cpu")

print("LOADED GPU")
```

What I get is:
```
LOADED CPU
Traceback (most recent call last):
  File "/home/energy/polfeb/Au_H2O/QMMM/training/lightning_logs/DELTA_MACE+GRAPH2MAT/version_0/checkpoints/load.py", line 7, in <module>
    cpt = torch.load("test_gpu.ckpt", map_location="cpu")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/energy/polfeb/myenv/lib/python3.11/site-packages/torch/serialization.py", line 1025, in load
    return _load(opened_zipfile,
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/energy/polfeb/myenv/lib/python3.11/site-packages/torch/serialization.py", line 1446, in _load
    result = unpickler.load()
             ^^^^^^^^^^^^^^^^
  File "/home/energy/polfeb/myenv/lib/python3.11/site-packages/e3nn/util/codegen/_mixin.py", line 109, in __setstate__
    smod = torch.jit.load(buffer)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/energy/polfeb/myenv/lib/python3.11/site-packages/torch/jit/_serialization.py", line 160, in load
    cpp_module = torch._C.import_ir_module_from_buffer(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx

```

**Expected behavior**
I should be able to load to CPU regardless of where the tensor product was when it was stored.

**Environment (please complete the following information):**
 - OS: Linux
 - python version Python 3.11.3
 - python environment (commands are given for python interpreter):
   - e3nn version: 0.4.4 (in this particular test, but it happens for the newest version as well)
   - pytorch version 2.3.0+cu121
   - pytorch_geometric version 2.5.3

pfebrer:
Fixes #461 

I added a test for the problem, but I don't know if it is a good idea since it needs a file to load. The most minimal example I could come up with results in a 26KB file.

Interestingly, I found that:

```python
from e3nn import o3
import torch

tp = o3.FullyConnectedTensorProduct("1x0e+1x1o", "1x0e+1x1o", "1x0e+1x1o+1x2e")
tp = tp.to("cuda")
torch.save(tp, "tp_gpu.ckpt")
```

creates issues, but removing `1x2e` from the `irreps_out` doesn't. Perhaps this gives you an idea of what could be going wrong internally, and you can come up with a better way to solve it.


pfebrer:
Hi @mitkotak, what do you need to merge this?

mitkotak:
Thanks for the PR ! The fact that it works with `1x2e` is still confusing to me but regardless if you can update the branch to match with main, I can trigger the tests and we can merge it in.  Sorry for the delay.

mitkotak:
Ok had a chat with Alby (who is much more familiar with this part of the codebase) and seems like this fix will be a band aid over the larger torch issue:


"
if you want a portable TorchScript archive (or any torch archive) you should move to CPU before saving you can always re-save it in the terminal with a `a = torch.jit.load("f.pth", map_location='cpu'); torch.jit.save("f.pth", a)` 

Then after loading you should move the loaded module / dict to target device, which is all that torch.jit.load is doing anyway:

> All previously saved modules, no matter their device, are first loaded onto CPU, and then are moved to the devices they were saved from.

https://pytorch.org/docs/stable/generated/torch.jit.load.html

It works without `2e `because it will use the "specialized code" paths for vector TPs that don't have w3js, which means there are no buffers

I think there may be some caveats to this (https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-model-across-devices?)

but this PR is not right way to solve them... the bigger problem is that the internal `__setstate__` `torch.jit.load` isn't getting propagated through any map_location that is set when calling torch.load of a whole pickled model. (https://pytorch.org/docs/stable/generated/torch.load.html)

Note that pytorch themselves encourage saving state dicts, and not pickling whole models: https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-entire-model

the entire ability to pickle `CodeGenMixin'd `classes is a bit of a best-effort kludge, mostly designed to enable `deepcopying` them

I would not encourage saving them as model checkpoints instead of just state dicts, which are far more portable and inspectable
"

mitkotak:
Ok had a chat with Alby (who is much more familiar with this part of the codebase) and seems like this fix will be a band aid over the larger torch issue:

"
if you want a portable TorchScript archive (or any torch archive) you should move to CPU before saving you can always re-save it in the terminal with a `a = torch.jit.load("f.pth", map_location='cpu'); torch.jit.save("f.pth", a)`

Then after loading you should move the loaded module / dict to target device, which is all that torch.jit.load is doing anyway:

> All previously saved modules, no matter their device, are first loaded onto CPU, and then are moved to the devices they were saved from.

https://pytorch.org/docs/stable/generated/torch.jit.load.html

It works without 2e because it will use the "specialized code" paths for vector TPs that don't have w3js, which means there are no buffers

I think there may be some caveats to this (https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-model-across-devices?)
but this PR is not right way to solve them... the bigger problem is that the internal `__setstate__` `torch.jit.load` isn't getting propagated through any map_location that is set when calling torch.load of a whole pickled model. (https://pytorch.org/docs/stable/generated/torch.load.html)

Note that pytorch themselves encourage saving state dicts, and not pickling whole models: https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-entire-model

the entire ability to pickle `CodeGenMixin'd` classes is a bit of a best-effort kludge, mostly designed to enable `deepcopying `them

I would not encourage saving them as model checkpoints instead of just state dicts, which are far more portable and inspectable
"

pfebrer:
For training and inference I use `pytorch_lightning` which saves and loads checkpoints, and I'm not sure I can modify that behavior. I'm also not sure that I can tell lightning to move the model to cpu before storing the checkpoint (or whether that would be wise, because it would need to transfer everything on each epoch).

I understand your points about `torch` best practices (in which I'm not very well versed) but in practical terms this issue will be a problem for anyone who uses `pytorch_lightning` or any other framework that stores checkpoints.

Given that the fix is as simple (and logic) as passing the map location, I think it would be worth it to support these "not so good practices", be it through this PR or any other way of solving it :)

pfebrer:
For a bit more motivation (in case that you are thinking that I want to do weird stuff), here is my use case: 

My ML models try to accelerate Density Functional Theory (DFT) simulations, where codes are quite efficient in running in parallel in multiple CPUs. It is not so clear that using GPUs is an advantage vs simply using more CPUs in terms of cost. I run my ML model hand in hand with a DFT simulation, and the cost of the model even in CPU is often negligible. It would therefore be a waste of money/resources to run the model on GPU. But the models are trained on GPU, therefore I get this problem with `e3nn`.

Linux-cpp-lisp:
Can you elaborate on how you are actually saving a checkpoint?  It seems like Lightning should just be saving the state dict in the checkpoint: https://lightning.ai/docs/pytorch/stable/common/checkpointing_basic.html#contents-of-a-checkpoint/

pfebrer:
Hmm yes now that you say it I seem to recall that lightning's checkpoint simply contains the `state_dict` of the model and the model is loaded with `load_state_dict`.

I don't manipulate the way in which lightning stores the checkpoints. 

The last 3 months I've been on holidays so the exact problem in lightning is a bit blurry right now in my mind. I'll try to run things again next week and get back to you with more details :+1:

Linux-cpp-lisp:
Sounds good, and thank you for checking!  While I agree that the changes here are small and it's important for things to work with a common framework like Lightning, I think it's important to be careful of putting band-aids that add complexity where future bugs can hide that do not resolve core reasons why there are issues.

pfebrer:
Ok so indeed I had modified a bit the checkpoint storing procedure and `lightning` was storing some tensor products directly by pickling them. This was unnecessary and now that I reverted the modification it seems that things are fine.

So I'll close this PR and the issue, thanks :+1:

PR #484, opened 2024-10-31
Force `CodeGenMixin` to entirely respect `e3nn.set_optimization_defaults`

Linux-cpp-lisp:


e3nn/o3/_tensor_product/_codegen.py
```diff
@@ -39,14 +39,20 @@ def codegen_tensor_product_left_right(
     x2s = fx.Proxy(graph.placeholder("x2", torch.Tensor), tracer=tracer)
     weights = fx.Proxy(graph.placeholder("w", torch.Tensor), tracer=tracer)
 
-    empty = fx.Proxy(graph.call_function(torch.empty, ((),), dict(device="cpu")), tracer=tracer)
+    # Short circuit for zero-dimensional inputs
+    if len(irreps_in1) == 0 or len(irreps_in2) == 0:
+        if shared_weights:
+            output_shape = torch.broadcast_tensors(x1s[..., :1], x2s[..., :1])[0].shape[:-1]
```
Linux-cpp-lisp, line 45: @mitkotak would this `:1` instead of `0` trick work if just applied below without even including the short circuit?

mitkotak:
@Linux-cpp-lisp Can you do a quick skim before I merge ?

e3nn/o3/_tensor_product/_codegen.py
```diff
@@ -39,16 +39,11 @@ def codegen_tensor_product_left_right(
     x2s = fx.Proxy(graph.placeholder("x2", torch.Tensor), tracer=tracer)
     weights = fx.Proxy(graph.placeholder("w", torch.Tensor), tracer=tracer)
 
-    empty = fx.Proxy(graph.call_function(torch.empty, ((),), dict(device="cpu")), tracer=tracer)
     if shared_weights:
-        output_shape = torch.broadcast_tensors(empty.expand(x1s.shape[:-1]), empty.expand(x2s.shape[:-1]))[0].shape
+        output_shape = torch.broadcast_tensors(x1s[..., :1], x2s[..., :1])[0].shape[:-1]
```
Linux-cpp-lisp, line 43:
```suggestion
        # by broadcasting all but the final irrep dim, we broadcast any and all batch dimensions
        # the use of `:1` is important, rather than `0`, to ensure the case with an empty irrep dimension doesn't error out
        output_shape = torch.broadcast_tensors(x1s[..., :1], x2s[..., :1])[0].shape[:-1]
```

e3nn/o3/_tensor_product/_codegen.py
```diff
@@ -415,12 +410,10 @@ def codegen_tensor_product_right(
     x2s = fx.Proxy(graph.placeholder("x2", torch.Tensor), tracer=tracer)
     weights = fx.Proxy(graph.placeholder("w", torch.Tensor), tracer=tracer)
 
-    empty = fx.Proxy(graph.call_function(torch.empty, ((),), dict(device="cpu")), tracer=tracer)
     if shared_weights:
         output_shape = x2s.shape[:-1]
     else:
-        output_shape = torch.broadcast_tensors(empty.expand(x2s.shape[:-1]), empty.expand(weights.shape[:-1]))[0].shape
-    del empty
+        output_shape = torch.broadcast_tensors(x2s[..., 0], weights[..., 0])[0].shape
```
Linux-cpp-lisp, line 416:
```suggestion
        # see broadcasting notes above
        output_shape = torch.broadcast_tensors(x2s[..., :1], weights[..., :1])[0].shape
```
should this also be adjusted?

PR #485, opened 2024-11-01
hack to turn off jit script in _spherical_harmonics

mitkotak:


tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -159,3 +160,15 @@ def build_module(l, normalize, normalization):
...
     sp_pt2 = torch.compile(prepare(build_module)(l, normalize, normalization), fullgraph=True)
     assert torch.allclose(sp_pt2(xyz), sp(xyz))
+
+
+def test_internal_jit_flag():
+    l = o3.Irreps("0e + 1o + 3o")
+    sp = o3.SphericalHarmonics(l, normalization="integral", normalize=True)
+    # Turning off the JIT in _spherical_harmonics to enable torch.compile.
+    # Note that this is a less invasive way then turning off jit_script_fx globally
+    set_optimization_defaults(jit_script_fx=False)
```
Linux-cpp-lisp, line 170: Need to do a context manager with a `try: finally` for this so it doesn't apply to all tests after.

Actually would not be bad to offer `set_optimization_defaults` as a context manager as well in `e3nn`

e3nn/o3/_spherical_harmonics.py
```diff
@@ -84,12 +85,20 @@ def __init__(
                 f"spherical_harmonics maximum l implemented is {_lmax}, send us an email to ask for more"
             )
 
+        # TODO: Using this to turn off the torch.jit.script for torch.compile.
+        # Please remove this when torch.script is depracated
+        if get_optimization_defaults()["jit_script_fx"]:
+            self.sph_func = torch.jit.script(_spherical_harmonics)
+        else:
+            # Turn off for torch.compile
+            self.sph_func = _spherical_harmonics
```
Linux-cpp-lisp, line 94: I am not sure if this will be picklable since it's a function and `ScriptFunction` may not be picklable, is that included in the tests for sh?

tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -156,6 +157,38 @@ def build_module(l, normalize, normalization):
...
     assert torch.allclose(sp_pt2(xyz), sp(xyz))
+
+
+def test_internal_jit_flag():
+    l = o3.Irreps("0e + 1o + 3o")
+    sp = o3.SphericalHarmonics(l, normalization="integral", normalize=True)
+    # Turning off the JIT in _spherical_harmonics to enable torch.compile.
+    # Note that this is a less invasive way then turning off jit_script_fx globally
+    try:
+        torch._dynamo.reset()
```
Linux-cpp-lisp, line 171: Why is this necessary?

tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -156,6 +157,38 @@ def build_module(l, normalize, normalization):
...
+        set_optimization_defaults(jit_script_fx=False)
+        sp_pt2 = torch.compile(o3.SphericalHarmonics(l, normalization="integral", normalize=True), fullgraph=True)
+
+        xyz = torch.randn(11, 3)
+        torch.testing.assert_close(sp(xyz), sp_pt2(xyz))
+    finally:
+        set_optimization_defaults(jit_script_fx=True)
```
Linux-cpp-lisp, line 178: This should be a true context manager that restores the previous state if we don't want it to break future tests where sometimes fx is off

tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -156,6 +157,38 @@ def build_module(l, normalize, normalization):
...
+
+
+def test_picke():
+    l = o3.Irreps("0e + 1o + 3o")
+    sp = o3.SphericalHarmonics(l, normalization="integral", normalize=True)
+    try:
+        torch._dynamo.reset()
```
Linux-cpp-lisp, line 185: same question

tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -156,6 +157,38 @@ def build_module(l, normalize, normalization):
...
+        set_optimization_defaults(jit_script_fx=False)
```
Linux-cpp-lisp, line 186: dont we need to test it with it on, or at least both ways?

tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -156,6 +157,38 @@ def build_module(l, normalize, normalization):
...
+        sp_pt2 = torch.compile(o3.SphericalHarmonics(l, normalization="integral", normalize=True), fullgraph=True)
+
+        import io
```
Linux-cpp-lisp, line 189: import at the top?

tests/o3/cartesian_spherical_harmonics_test.py
```diff
@@ -156,6 +157,38 @@ def build_module(l, normalize, normalization):
...
+
+        buffer = io.BytesIO()
+        torch.save(sp_pt2.state_dict(), buffer)
```
Linux-cpp-lisp, line 192:
```suggestion
        torch.save(sp_pt2, buffer)
```

State dict will always work, needs to be the module itself

tests/o3/cartesian_spherical_harmonics_test.py
mitkotak: The compiler cache gets built up from previous runs so need to flesh that out

mitkotak: Looks like bumping to 2.2.0 -> 2.4.0 fixed that one

e3nn/o3/_spherical_harmonics.py
mitkotak: ```jit_script_fx=True``` does seem to be unpickable so open to any suggestions that you might have 
https://github.com/pytorch/pytorch/issues/61210

PR #489, opened 2024-11-06
PyPI release 0.5.4

mitkotak:


SiLiKhon:
Hi! I have this version frozen in my `requirements.txt`, but it is now missing from pypi. Was it removed intentionally?

mitkotak:
Sorry for that ! PyPI shenanigans 

Was trying to publish a minor release for 0.5.4.1 but PyPI was not cooperating so tried removing it with the hope that the publish API would work. Eventually had to go for 0.5.5. 

Hopefully was not too much of an inconvenience

SiLiKhon:
thanks a lot @mitkotak for the explanation and the prompt response!

No problem. I just got worried if 0.5.4 has been retracted for some critical problem or something...
Glad to hear it's nothing like that :)

PR #497, opened 2024-12-12
[WIP] Compile on by default

mitkotak:
Currently to enable full `torch.compile` capability we need to do `e3nn.set_optimization_defaults(jit_script_fx=False)`.  I am proposing to turn this off by default.

Update: 01/03/2024

- New interface will have ``"jit_mode": "inductor/script/eager"`` 

- Will retain backward compatibility by mapping ``"jit_script_fx"=False`` to ``"jit_mode": "eager"`` and ``"jit_script_fx"=True`` to ``"jit_mode": "script"`` (Will only do this if ``"jit_mode"`` not in ``"inductor/script"``.

### DIFF
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -57,12 +57,14 @@ def set_optimization_defaults(**kwargs) -> None:
         if k not in _OPT_DEFAULTS:
             raise ValueError(f"Unknown optimization option: {k}")
 
-        # Handle the legacy jit_script_fx flag
+        # Handles the legacy mapping for jit_script_fx
+        # to jit_mode so that old code can still work
+        # with the new defaults.
         if k == "jit_script_fx":
-            _OPT_DEFAULTS[k] = v
             # Update jit_mode based on the legacy mapping
             new_jit_mode = _handle_jit_script_fx_legacy(v, _OPT_DEFAULTS["jit_mode"])
             _validate_and_set_jit_mode(new_jit_mode)
+            _OPT_DEFAULTS[k] = v
         elif k == "jit_mode":
             # Validate and set the new jit_mode
             _validate_and_set_jit_mode(v)
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -7,7 +7,7 @@ from typing import Dict
 _OPT_DEFAULTS: Dict[str, bool] = dict(
     specialized_code=True,
     optimize_einsums=True,
-    jit_script_fx=False,
+    jit_script_fx=True,
 )
 
 
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -8,7 +8,7 @@ _OPT_DEFAULTS: Dict[str, bool] = dict(
     specialized_code=True,
     optimize_einsums=True,
     jit_script_fx=True,
-    jit_mode="compile"
+    jit_mode="script"
 )
 
 def _handle_jit_script_fx_legacy(jit_script_fx: bool, current_jit_mode: str) -> str:
--- a/e3nn/__init__.py
+++ b/e3nn/__init__.py
@@ -8,8 +8,35 @@ _OPT_DEFAULTS: Dict[str, bool] = dict(
     specialized_code=True,
     optimize_einsums=True,
     jit_script_fx=True,
+    jit_mode="compile"
 )
 
+def _handle_jit_script_fx_legacy(jit_script_fx: bool, current_jit_mode: str) -> str:
+    """Handle the legacy jit_script_fx flag mapping to jit_mode.
+
+    Parameters
+    ----------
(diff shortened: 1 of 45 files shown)


mitkotak:
One annoying thing I am noticing is that if we do

```python

jitted = assert_auto_jitable(model)

pt2 = torch.compile(model, fullgraph=True)
```

then it tries to compiles a scripted module compared to 

```python
pt2 = torch.compile(model, fullgraph=True)

jitted = assert_auto_jitable(model)
```

which runs without errors

tests/nn/activation_test.py
```diff
@@ -15,24 +14,25 @@
...
 
-    a = build_module(irreps_in, acts)
-    a_pt2 = prepare(build_module)(irreps_in, acts)
     assert_auto_jitable(a)
     assert_equivariant(a)
 
-    inp = irreps_in.randn(13, -1)
     out = a(inp)
+
+    a_pt2 = torch.compile(a, fullgraph=True)
```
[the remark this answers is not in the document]
mitkotak: Thanks done !

e3nn/util/codegen/_mixin.py
```diff
@@ -39,11 +39,13 @@ def _codegen_register(
...
             assert isinstance(graphmod, fx.GraphModule)
 
-            if opt_defaults["jit_script_fx"]:
+            if opt_defaults["jit_mode"] == "script":
                 # With recurse=False, this more or less is equivalent to
                 # torch.jit.script(jitable(graphmod))
                 scriptmod = e3nn.util.jit.compile(graphmod, recurse=False)
                 assert isinstance(scriptmod, torch.jit.ScriptModule)
+            elif opt_defaults["jit_mode"] == "compile":
+                scriptmod = torch.compile(graphmod, fullgraph=True)
```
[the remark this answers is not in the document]
mitkotak: Got rid of the branch

e3nn/util/codegen/_mixin.py
```diff
@@ -76,7 +76,7 @@ def __getstate__(self):
                 smod = getattr(self, fname)
                 buffer_type: str
                 buffer: bytes
-                if isinstance(smod, fx.GraphModule):
+                if isinstance(smod, (fx.GraphModule, torch._dynamo.OptimizedModule)):
```
Linux-cpp-lisp, line 79: Are these `torch._dynamo.OptimizedModule` actually picklable?  Only within the same process?

e3nn/__init__.py
```diff
@@ -22,7 +56,18 @@ def set_optimization_defaults(**kwargs) -> None:
     for k, v in kwargs.items():
         if k not in _OPT_DEFAULTS:
             raise ValueError(f"Unknown optimization option: {k}")
-        _OPT_DEFAULTS[k] = v
+
+        # Handle the legacy jit_script_fx flag
+        if k == "jit_script_fx":
+            _OPT_DEFAULTS[k] = v
```
Linux-cpp-lisp, line 62: Should this happen after the validation?

More generally, is the rationale for this to make sure that older code that looks at this flag can still find it in the defaults dict? Maybe worth documenting that with a comment if so

e3nn/util/codegen/_mixin.py
mitkotak: Yup was able to confirm with a reproducer from Claude


```python
import torch
import pickle
import io

# First, let's create a simple PyTorch model
class SimpleModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = torch.nn.Linear(10, 5)
    
    def forward(self, x):
        return self.linear(x)

# Create an instance of the model
model = SimpleModel()

# Now, let's optimize the model with torch.compile
# This should wrap it in an OptimizedModule
optimized_model = torch.compile(model)

# Check the type of the optimized model
print(f"Type of optimized model: {type(optimized_model)}")
print(f"Is it an OptimizedModule? {isinstance(optimized_model, torch._dynamo.OptimizedModule)}")

# Try to pickle the optimized model
try:
    # Use a BytesIO object to pickle in memory
    buffer = io.BytesIO()
    pickle.dump(optimized_model, buffer)
    buffer.seek(0)
    
    # Try to unpickle
    unpickled_model = pickle.load(buffer)
    
    print("Successfully pickled and unpickled the OptimizedModule!")
    print(f"Type of unpickled model: {type(unpickled_model)}")
    
    # Check if the unpickled model works
    test_input = torch.randn(2, 10)
    original_output = optimized_model(test_input)
    unpickled_output = unpickled_model(test_input)
    
    # Compare outputs
    print(f"Outputs match: {torch.allclose(original_output, unpickled_output)}")
    
except Exception as e:
    print(f"Failed to pickle/unpickle: {e}")
    print("torch._dynamo.OptimizedModule is NOT picklable")
```

PR #505, opened 2025-04-02
Support for pre-computed Wigner tensors

mitkotak:


e3nn/o3/_wigner.py
```diff
@@ -1,13 +1,75 @@
...
+    Returns
+    -------
+    `torch.Tensor`
+        tensor :math:`D^l(\alpha, \beta, \gamma)` of shape :math:`(2l+1, 2l+1)`
+    """
+    _Jd = get_jd()
+    assert l < len(_Jd), f"wigner D maximum l implemented is {len(_Jd) - 1}, send us an email to ask for more"
+    # Currently goes till lmax=10
+    alpha, beta, gamma = torch.broadcast_tensors(alpha, beta, gamma)
+    J = _Jd[l].to(dtype=alpha.dtype, device=alpha.device)
```
Linux-cpp-lisp, line 67: This needs a `copy=True` to be safe

PR #512, opened 2025-07-21
Update _activation.py

changzhiai:
Use zip(self.irreps_in, self.acts) rather than self.paths to avoid torchscript error. 

## Description
In e3nn.nn._activation.py in line 98, `for mul, (l, _), act in self.paths:` can work well under pytorch but there would be an in compatible error within torchscript environment. To fix this, `for (mul, (l, _)), act in zip(self.irreps_in, self.acts): ` can work perfectly and realize exactly the same function.

## Motivation and Context
Fix torchscript incompatible error.

## How Has This Been Tested?
Yes, it works perfectly after the patch.

## Checklist:

### DIFF
--- a/e3nn/nn/_activation.py
+++ b/e3nn/nn/_activation.py
@@ -95,7 +95,8 @@ class Activation(torch.nn.Module):
         # - PROFILER - with torch.autograd.profiler.record_function(repr(self)):
         output = []
         index = 0
-        for mul, (l, _), act in self.paths:
+        # for mul, (l, _), act in self.paths:
+        for (mul, (l, _)), act in zip(self.irreps_in, self.acts): # Fix torchscript incompatible error
             ir_dim = 2 * l + 1
             if act is not None:
                 output.append(act(features.narrow(dim, index, mul)))


e3nn/nn/_activation.py
```diff
@@ -95,7 +95,8 @@ def forward(self, features, dim: int = -1):
         # - PROFILER - with torch.autograd.profiler.record_function(repr(self)):
         output = []
         index = 0
-        for mul, (l, _), act in self.paths:
+        # for mul, (l, _), act in self.paths:
```
mitkotak, line 98: Thanks ! Can you delete the commented out line instead of commenting it for clarity ?

mitkotak:
Sorry what I meant was that we need to support both torch.compile and script so was thinking of moving some of the logic into init but for that I first wanted to have a test case for script that replicates your setup

changzhiai:
Sorry about the pytest errors. Now I do the pytest check by myself, and errors should be fixed now, which is for `script` mode.

I should be more clear about this. The reason is I would like to use `script` compile mode rather than `trace` mode, where I found `script` mode is more flexible when I try to compile ML mode to torchscript. 

Regarding the pytest errors, they are from this line `assert_torch_compile("inductor", functools.partial(Activation, irreps_in, acts), inp)` in activation_test.py, which is to assert if it can work under torch.compile(fullgraph=True). It is not necessary and even break things in `script` mode. So I removed this line in activation_test.py.

I hope it is reasonable for you. 
Thanks a lot for your comments!

PR #517, opened 2025-09-15
Remove for loop over multiplicities in D_from_angles

SimonWagnerHD:
Remove the unnecessary for loop over multiplicities when using the function D_from_angles in the Irreps class.

## Description

Line 743 in _irreps.py `direct_sum(*[ir.D_from_angles(alpha, beta, gamma, k) for mul, ir in self for _ in range(mul)])` performs a for loop over multiplicities which in each iteration calculates `ir.D_from_angles` even though the result is the same each time. The suggested change calculates `D_from_angles` only once per irrep and repeats the resulting block according to the multiplicity.

## Motivation and Context

The change significantly increases the speed of the function especially in scenarios, where the number of multiplicities is high.

## How Has This Been Tested?

The changes were tested by comparing the results of old and new implementation on an irreps object `Irreps("128x0e + 64x1o + 16x2e + 4x3o + 1x4e")` for random tensors of angles and k and asserting that the resulting tensors are the same.

## Checklist:

### DIFF
--- a/e3nn/o3/_irreps.py
+++ b/e3nn/o3/_irreps.py
@@ -57,10 +57,10 @@ class Irrep(tuple):
         if p is None:
             if isinstance(l, Irrep):
                 return l
-            
+
             if isinstance(l, _MulIr):
                 return l.ir.l
-            
+
             if isinstance(l, str):
                 try:
                     name = l.strip()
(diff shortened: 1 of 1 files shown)


SimonWagnerHD:
I tested the runtime of new and old version on a NVIDIA GeForce RTX 2080 Ti. I used the same random tensors of size 1000 for the angles while changing the irreps according to "{m}x0e+{m}x1o+{m}x2e" with m in [1, 2, 4, 8, 16, 32, 64].

gpu_runtime

mitkotak:
Thanks for the PR ! Should be available in the latest release 0.5.8

PR #535, opened 2026-03-29
Fix loading of constants in _wigner.py to include weights_only=False …

bluehope:
Description
Add weights_only=False to torch.load() call in e3nn/o3/_wigner.py.

PyTorch 2.6 changed the default of `torch.load from weights_only=False` to `weights_only=True`. Since constants.pt contains a dict (_W3j_indices), loading fails without this explicit parameter.

Motivation and Context
import e3nn raises an error on PyTorch >= 2.6 due to the weights_only default change. This is a one-line fix to restore compatibility.

This issue only affects v0.4.4 and earlier, as constants.pt was removed in v0.6.0. However, widely used packages such as MACE still depend on e3nn 0.4.4, so it would be helpful to have this fix available upstream pytorch.

How Has This Been Tested
Verified import e3nn succeeds with PyTorch 2.6+ after the patch.
Confirmed e3nn.o3.wigner_3j and e3nn.o3.wigner_D return correct values.
Checklist
 I have read the [CONTRIBUTING](https://github.com/e3nn/e3nn/blob/main/.github/CONTRIBUTING.md) document.
 My code follows the code style of this project.
 I have updated the documentation (if relevant).
 I have added tests that cover my changes (if relevant).
 The modified code is cuda compatible (github tests don't test cuda) (if relevant).
 I have updated the [CHANGELOG](https://github.com/e3nn/e3nn/blob/main/.github/CHANGELOG.md).


mitkotak:
@bluehope Thanks for the PR ! I am seeing conflict with the main branch on my side. Do you also see it ? The docs are also not building for some reason.

bluehope:
Hi @mitkotak , thanks for reviewing!
Oh, just to clarify the intent: 
this PR was primarily aimed at v0.4.4, not main. 
The fix is needed because packages like MACE still depend on e3nn 0.4.4, and PyTorch 2.6+ breaks the torch.load() call in that version. Since v0.6.0 already removed the constants file, this issue doesn't exist on main — which explains the conflicts you're seeing.

I already have the fix ready on a branch based on the v0.4.4 tag in my fork: [KNU-CMS/e3nn/tree/v0.4.4-pt25fix](https://github.com/KNU-CMS/e3nn/tree/v0.4.4-pt25fix)
Could you either:

Create a maintenance branch from the v0.4.4 tag so I can retarget this PR properly, or
Cherry-pick the fix directly and cut a v0.4.5 patch release?

Given how many downstream packages are affected, a patch release would be really valuable for the ecosystem. Thanks!

PR #536, opened 2026-07-14
Add Cartesian operations, for Cartesian and SO(2) basis

xvzemin:
## Description
This PR adds a new `e3nn.o3._zemin` module with utilities for Irreducible Cartesian tensor:

- `ICTD` performs irreducible Cartesian tensor decomposition. It establishes an explicit correspondence between Cartesian and spherical tensors and provides the associated change-of-basis transformations.
- `cartesian_3j`, which constructs a Cartesian analogue of the Wigner-3j coefficients by contracting Wigner-3j coefficients with highest-weight path matrices. With suitable modifications, it can also be used to directly couple irreducible Cartesian tensors (ICTs) with spherical tensors (STs).
- `CartesianHarmonics`, which builds ICT from vector.
- `CartesianWignerD`, which computes block-diagonal Wigner-D matrices from relations between ICT and ST using either a `recursive` or `direct` method. The module takes rotation matrices directly as input, enabling higher numerical accuracy while avoiding the discontinuities encountered in eSCN and the need for the specialized quaternion kernels used in UMA.

The new APIs are exported from `e3nn.o3.__init__`.
## Motivation and Context
This adds  Cartesian tensor functionality to `o3` without modifying existing e3nn core implementations. 

## How Has This Been Tested?
Added `tests/o3/zemin_test.py` covering:

- ICTD path orthonormality and completeness.
- `cartesian_3j` output shape and normalization.
- `CartesianHarmonics` output dimensions for spherical-style `Irreps`, including non-contiguous requested degrees.
- Tracelessness of rank-2 Cartesian harmonics.
- Agreement among the `recursive` and `direct` methods of `CartesianWignerD` and `e3nn.o3.wigner_D`. The recursive method generally provides higher numerical accuracy, so differences in numerical precision should be taken into account when comparing it with the native implementation.
Test command:
`python -m pytest tests/o3/zemin_test.py -q`

## Checklist:


mitkotak:
Hey thanks for the PR. Nice to see you again ! Just getting down to reviewing it. Can you also add docs + examples (can be a simple network from your main repo).

mitkotak:
Ok instead of introducing a completely new module how much can be done with ReducedTensorProducts  as done in _cartesian_tensor.py ?

xvzemin:
Thanks! Nice to see you again too.

# ReducedTensorProducts and ICTD

About `ReducedTensorProducts`: after looking at it more carefully, I agree that it can already provide the complete change of basis for Cartesian tensors, and in that sense it overlaps with ICTD.

The `change_of_basis` constructed by ICTD and `ReducedTensorProducts` appear to differ by the additional `yzx`-to-`xyz` change of basis described in the e3nn documentation.

And the ICTD utility here exposes the decomposition in a more Cartesian-native and path-resolved form, by returning explicit parentage-path matrices and projectors.

My preference is to keep the explicit ICTD interface available, as it provides an explicit interface and a simpler implementation.

# Naive Complete O(3) Irreducible Cartesian tensor product and contraction

As for whether this should be added as a new module, my original intention was to add an implementation of the polar and pseudo branch for native irreducible Cartesian tensor product/contraction (ICTP/ICTC) functionality when time permits. Since this is a relatively substantial change and cannot be implemented entirely by directly reusing the existing e3nn components, I created a separate module and temporarily gave it a name ending with my own name. The general implementation method is

## Polar branch
```math
( A \otimes_k^\delta B)_{i j}
=
\frac{1}{\sqrt{3^k}}
\sum_a
A_{i a}
B_{a j},
```

where `k` axes are contracted, and

```math
\ell_3 = \ell_1 + \ell_2 - 2k,
\qquad
0 \le k \le \min(\ell_1,\ell_2).
```

The corresponding irreducible Cartesian output is obtained by applying the ICTD projector:

```math
T_{\mathrm{out}}^{(\ell_3;\ell_3;1)}
=
ICTD^{(\ell_3;\ell_3;1)}
( A \otimes_k^\delta B).
```

Thus `k=0` gives the highest-weight ICTP and `k>0` gives the highest-weight ICTC.

## Pseudo Branch

The pseudo branch uses one Levi-Civita tensor:

```math
( A \otimes_k^\epsilon B)_{i r j}
=
\frac{1}{\sqrt{2}}\frac{1}{\sqrt{3^k}}
\sum_a\sum_{b,c}
A_{i b a}
\epsilon_{rbc}
B_{a c j},
```

with

```math
\ell_3 = \ell_1+\ell_2-2k-1,
\qquad
0 \le k < \min(\ell_1,\ell_2).
```

The corresponding irreducible Cartesian output is obtained by applying the ICTD projector:

```math
T_{\mathrm{out}}^{(\ell_3;\ell_3;1)}
=
ICTD^{(\ell_3;\ell_3;1)}
( A \otimes_k^\epsilon B).
```

# Regarding the documentation and examples, 

Based on your feedback, I’ll decide whether the code needs to be revised. The documentation and example will be added over the next few days.

It would also be helpful if you could assess whether these four components are appropriate additions to e3nn:

* `ICTD`
* `CartesianHarmonics`
* `Cartesian 3j`
* `CartesianWignerD`

Among them, I think `CartesianWignerD` currently provides the most practical value. Its recursive construction is relatively simple, offers better numerical accuracy than constructing the transformation through Euler angles, and does not require a specialized kernel like quarternion. For generality, the current implementation does not explicitly expose an `mmax` truncation option for the SO(2) case, although this could be added if it would be useful.

xvzemin:
We plan to develop full support for O(2) and Cartesian O(3), so we will close this PR for now. If there is an opportunity to revisit this in the future, we will open a new PR. Thank you!

